PR #224, opened 2015-01-07
Run apt-get update before installing packages

cap10morgan:
This helps prevent the missing packages problem. Still not a perfect
solution, of course, but hey, it builds now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,12 +6,11 @@
 FROM quay.io/democracyworks/clojure-api-supervisor:latest
 MAINTAINER Democracy Works, Inc. <dev@democracy.works>
 
-# install MongoDB
-RUN apt-get install -y mongodb
-RUN mkdir -p /data/db
+# install some dependent packages
+RUN apt-get update && apt-get install -y mongodb git build-essential libfreetype6 libfontconfig
 
-# install git (bower dependency)
-RUN apt-get install -y git
+# create MongoDB data dir
+RUN mkdir -p /data/db
 
 # install Node
 RUN add-apt-repository -y ppa:chris-lea/node.js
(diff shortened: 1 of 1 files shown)


cap10morgan:
For @tie-rack in case it's helpful.

tie-rack:
It may build, but it doesn't run.

```
exec: "/run.sh": stat /run.sh: no such file or directoryFATA[0000] Error respons
e from daemon: Cannot start container a72fa7ea42998e6384f3039345f63af62e6bd55cf1
d7d1841d28f274b726e9ee: exec: "/run.sh": stat /run.sh: no such file or directory 
```

cap10morgan:
You didn't say anything about *running* it, sheesh. That's probably just good old fashioned bit rot. I don't think anyone has used this Dockerfile until you tried.

tie-rack:
Maybe I'm missing something, but it doesn't look like this docker container was *ever* runnable. The Dockerfile has *always* referenced a `run.sh` file which has *never* existed.

And if no one's used this Dockerfile before, why is it here? Why is it at the top of the README? To what end has it been maintained? Should it be removed? Or fixed? Or started over?

cap10morgan:
Honestly it's been so long since I've touched it I don't remember the relevant details. I would say we should fix it.

cap10morgan:
To address your other questions about why there's a Dockerfile at all: At some point we do want to run this and any other VIP services under a similar Docker hosting infrastructure as the one we settle on for TurboVote, Ballot Scout, etc.

The README stuff seems to have been added prematurely. I'm horrified that past me could have put that in there without actually trying it. But I guess I did. Mea culpa.

tie-rack:
You know, I've got this working. Part of it, I guess, is that we just need to update our base image (and then the intermediate images up to this one (which will also help papertrail).

I'll open a PR that gets this running in Docker. Closing for now.

PR #225, opened 2015-01-09
[CS] Fix running Metis inside docker.

tie-rack:
There is no "/run.sh". There is no CMD. Supervisor takes care of everything.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,4 +49,3 @@ RUN ln -s /metis/docker/run-tests.sh /run-tests.sh
 
 # run the app
 EXPOSE 4000 27017 28017
-CMD ["/run.sh"]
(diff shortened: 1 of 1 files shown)


tank157:
LGTM

tie-rack:
@jungshadow Metis never seems to have ever worked inside Docker. The erroneous `CMD` directive in the Dockerfile prohibited the container from running at all. After its removal, the container itself would start, but the version of MongoDB being installed from `apt` was an old version that doesn't support some functionality Metis requires.

The first commit took the Docker container from fundamentally unrunnable to runnable but partially broken. This new commit gets not only the container running, but Metis itself.

(Ideally, of course, you're not running your database and application in the same container. But one step at a time.)

PR #226, opened 2015-05-14
Dockerization

tank157:
Changes the Dockerization to extend from standard Dockerfiles, and runs Mongo out of a separate Dockerfile that also inserts the seed data for counties and state FIPS.

Adds docker-compose to simplify the linked running process, so with two simple commands you can have a running Metis.

[Pivotal Card 1](https://www.pivotaltracker.com/story/show/92966568)
[Pivotal Card 2](https://www.pivotaltracker.com/story/show/92966744)

Assigning to @tie-rack

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,9 +10,6 @@ MAINTAINER Democracy Works, Inc. <dev@democracy.works>
 RUN npm install -g grunt-cli
 RUN npm install -g bower
 
-# install build-essential for dependencies that need that stuff
-# RUN apt-get install -y build-essential
-
 # install bower deps (TODO: move this to cached step)
 RUN cd /usr/src/app && bower --allow-root install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,53 +3,18 @@
 # three (clojure-api, clojure-api-supervisor, and supervisor). One day Docker
 # will hopefully obviate the need for this with some shiny new INCLUDE command
 # or something similar. - WM 2014-7-29
-FROM quay.io/democracyworks/clojure-api-supervisor:latest
+FROM node:0.12.2-onbuild
 MAINTAINER Democracy Works, Inc. <dev@democracy.works>
-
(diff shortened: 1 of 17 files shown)


Dockerfile
```diff
@@ -3,53 +3,18 @@
...
-
-# add app code to container
-ADD . /metis
-
-# setup supervisor
-RUN chmod +x /metis/docker/*.sh
-RUN ln -s /metis/docker/supervisord-mongodb.conf /etc/supervisor/conf.d/
-RUN ln -s /metis/docker/supervisord-import-seed-data.conf /etc/supervisor/conf.d/
-RUN ln -s /metis/docker/supervisord-metis.conf /etc/supervisor/conf.d/
+# RUN apt-get install -y build-essential
```
tie-rack: If this isn't necessary, might as well delete it.

data/mongo/mongo.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+mongod --smallfiles &
+sleep 5
+mongoimport -d metis -c counties < data/counties.json
+mongoimport -d metis -c statefips < data/stateFIPS.json
+wait
```
tie-rack: Is it unfortunate that this script isn't usable outside of the mongo Docker? If the locations of the counties.json and stateFIPS.json files were supplied as arguments to this script (and the data/Dockerfile was updated to pass the Dockerized locations), this script could be used outside of docker.

Does that sound right? Something worth doing?

tank157: It's an enticing idea, but the mongo container we extend from adds certain directories that aren't typically available on a dev machine (i.e. /data/db) that running this script expects. You could also supply a mongod.conf file which then can specify another data location. Let me see if it can work without too much headache.

tank157:
@tie-rack your wish is granted

tie-rack:
Awesome. LGTM.

PR #227, opened 2015-05-19
Front end updates 1

bwreid:
This PR encompasses the following cards that update the front-end of Metis:

* [Pivotal Card 93958314](https://www.pivotaltracker.com/story/show/93958314): Remove Results link in the navigation, the Results table on the "Overview" page, and related code
* [Pivotal Card 93946948](https://www.pivotaltracker.com/story/show/93946948): Breadcrumbs on the "Overview" page should say something understandable about the upload being viewed
* [Pivotal Card 93947840](https://www.pivotaltracker.com/story/show/93947840): Add Election data to the Source page

Tests were updated to be in line with the above cards and the stateFIPS file has been changed slightly as it was not importing correctly.

To see these changes, start boot2docker and then run `docker-compose build && docker-compose up`.


dcohen:
LGTM.

PR #228, opened 2015-05-21
Front end changes 3

dcohen:
This PR includes the following cards:

<a href="https://www.pivotaltracker.com/n/projects/1106394/stories/93958330">Modify the red Total Errors in Feed warning box</a>

<a href="https://www.pivotaltracker.com/n/projects/1106394/stories/93958320">Add an "Errors" link on the navigation to show all errors for a feed</a>

The main functional change here is the "Download Full Error Report" button, which has moved into the sidebar and has a new Angular route associated with it - clicking that link should download a full CSV of all reported errors relative to a base feed URL path.

### DIFF
--- a/public/app/partials/aside.html
+++ b/public/app/partials/aside.html
@@ -26,6 +26,7 @@
         <li><a id="sidebar-election" class="is-finished" href="{{getAngularUrl(feedData.election)}}">Election</a></li>
         <li><a id="sidebar-pollinglocations" class="is-started" href="{{getAngularUrl(feedData.state)}}">Polling Locations</a></li>
         <li><a id="sidebar-contests" class="is-started" href="{{getAngularUrl(feedData.election_contests)}}">Contests</a></li>
+        <li><a id="sidebar-errors" class="is-started" href="{{getAngularUrl(feedData.errors)}}">Errors</a></li>
       </ul>
     </nav>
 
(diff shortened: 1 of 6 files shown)


public/app/partials/aside.html
```diff
@@ -10,11 +10,15 @@
   <div class="wrapper">
 
   <div class="aside-nav">
-    <a class="button-errors" href="{{getAngularUrl(feedData.errors)}}">
+    <span class="button-errors">
       {{feedData.error_count | number}} Total Errors in Feed
       <br>
       <time id="feedDueDate" ng-if="getDueDateText(feedData.date)" class="due-date"><i class="fi-clock"></i> Final feed due {{getDueDateText(feedData.date)}}</time>
-    </a>
+      <a class="button button-download" href="{{getUrlPath(feedData.errors)}}?error_report">Download Full Error Report</a>
```
bwreid: I don't think we need `getUrlPath` here, or the associated function created below. Instead, you could just have:
`{{getUrlPath(feedData.errors)}}?error_report`

bwreid:
I added an update to the Dockerfile to this PR to compile SCSS during the build process. Otherwise, the CSS wasn't being created.

After addressing my other comment, this looks good to me.

dcohen:
@bwreid OK, cool. I updated it to use that route instead and verified that the report data is the same. Should be good to go.

PR #230, opened 2015-05-26
Front end updates 4

dcohen:
This PR addresses <a href="https://www.pivotaltracker.com/n/projects/1106394/stories/93958188">this card</a> in Pivotal - this table previously existed on the overview page and needed to be added to the Contests page as well.


public/assets/js/app/controllers/feedContestsController.js
```diff
@@ -46,6 +48,29 @@ function FeedContestsCtrl_getFeedContests($scope, $rootScope, $feedsService, ser
   $feedsService.getFeedContests(servicePath)
     .success(function(data) {
 
+
+      // set the feeds data into the Angular model
+      $scope.feedContests = data;
```
bwreid: Does this not get overwritten by the function below? Do you see both tables when it's getting loaded?

dcohen:
Yes, they both load - see screenshot.
vip_dashboard

bwreid:
Alright cool. LGTM!  :)

PR #233, opened 2015-06-09
Pg sidebar

bwreid:
A solid start to connecting PostgreSQL & Metis.

[Pivotal Card](https://www.pivotaltracker.com/story/show/95368494)

This card allows for PostgreSQL & Metis to talk to each other. In this instance, we are solely returning the full count of errors from a valid run of data-processor. However, this also sets us up to be able to do further connections more easily.


pg/gets.js
```diff
@@ -0,0 +1,60 @@
+var conn = require('./conn.js');
+
+module.exports = {
+  // Functions below return arrays for the various queries with the requirement of an ID.
+  getResults: function(req, res) {
+    var client = conn.openPostgres();
+    var query = client.query("SELECT * FROM results WHERE id=" + req.query.id);
```
tie-rack: These all look like prime candidates for SQL injection. What if `req.query.id` is `NULL; DROP * FROM results`? The postgres library should have a way to build parameterized queries so that we're guarded from that situation.

pg/conn.js
```diff
@@ -0,0 +1,20 @@
...
+
+    var client = new pg.Client(connString);
+    client.connect();
+    return client;
+  },
+  // Close the connection
+  closePostgres: function(query, client, res) {
+    query.on("end", function (result) {
+      client.end();
+      res.writeHead(200, {'Content-Type': 'text/plain'});
```
tie-rack: If we're sending JSON back, should the content type be `application/json`?

pg/gets.js
```diff
@@ -0,0 +1,60 @@
...
+      
+      for (i = 0; i < result.rows.length; i++) {
+        var parsed_message = edn.toJS(edn.parse(result.rows[i]["message"]));
+        count += parsed_message.length || 1;
+      }
+
+      client.end();
+      res.writeHead(200, {'Content-Type': 'text/plain'});
+      res.write(JSON.stringify(count, null, "    ") + "\n");
+      res.end();
```
tie-rack: Perhaps this `res.writeHead... res.end()` block should be a function. It's repeated 3 times.

pg/gets.js
tie-rack: [Looks like](https://github.com/brianc/node-postgres/wiki/Client#parameterized-query-using-stringarray-initialization) this could be done like so:

```javascript
client.query("SELECT * FROM results WHERE id = $1", [req.query.id]);
```

And similarly for the rest.

tie-rack:
LGTM!

PR #234, opened 2015-06-09
[DC] updated test cases

dcohen:
These test cases have been updated to reflect the new breadcrumb structure - explicit calls to the "Feeds" link have been removed, and where links are being called by index number, the indices have been adjusted to reflect the fact that the "Feeds" link no longer exists at position 0.


bwreid:
This is great.

PR #236, opened 2015-06-15
Populate feed overview table from postgres

tie-rack:
Pivotal story: [95368418](https://www.pivotaltracker.com/story/show/95368418)

overview

Additionally, refactor some patterns that have emerged.

### DIFF
--- a/public/app/partials/feed-overview.html
+++ b/public/app/partials/feed-overview.html
@@ -13,8 +13,8 @@
   <section class="data-group data-module">
     <div ng-if="!overviewData"></br><span class="is-loading"></span></div>
 
-    <table class="overview" ng-show="overviewData.length" id="pollingTable" template-pagination="pollingPagination" ng-table="pollingTableParams" class="associations">
-      <tr id="pollingLocation{{$index}}" ng-repeat="pollingLocation in overviewData[0].pollingLocations">
+    <table class="overview" ng-show="overviewData" id="pollingTable" template-pagination="pollingPagination" ng-table="pollingTableParams" class="associations">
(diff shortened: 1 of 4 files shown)


jungshadow:
Quick unrelated question on this: why is the "Download Full Error Report" button blue? I don't think that was in the style guide (which I'm not sure is up-and-functioning anymore).

public/app/partials/feed-overview.html
```diff
@@ -2,39 +2,21 @@
...
   <!-- element table directive -->
   <section class="data-group data-module">
-    <div ng-if="!feedPollingLocations"></br><span class="is-loading"></span></div>
-    
-    <table class="overview" ng-show="feedPollingLocations.length" id="pollingTable" template-pagination="pollingPagination" ng-table="pollingTableParams" class="associations">
-      <tr id="pollingLocation{{$index}}" ng-repeat="pollingLocation in $data">
+    <div ng-if="!overviewData"></br><span class="is-loading"></span></div>
+
+    <table class="overview" ng-show="overviewData.length" id="pollingTable" template-pagination="pollingPagination" ng-table="pollingTableParams" class="associations">
+      <tr id="pollingLocation{{$index}}" ng-repeat="pollingLocation in overviewData[0].pollingLocations">
```
bwreid: Having overviewData[0] everywhere is a little meh. Could we maybe just redefine this in the controller to something?

dcohen:
It'll be red - the testing environment we're working in hasn't had the recent style changes applied to it. Running 'grunt sass' will update app.css and it'll look correct.

bwreid:
Looks Good To Me

PR #237, opened 2015-06-17
Use results table's public_id, not id

tie-rack:
Depends on votinginfoproject/data-processor#62.

URLs will use the "public_id" from the results table, so querying for data will need to go through the results table to get the results.id to join on.


bwreid:
Once the other is merged, LGTM.

PR #242, opened 2015-06-19
Contest pages

bwreid:
This PR:
* Updates the contests page to work with the new public_id
* Connects the individual contest page to postgres

Missing from this is connecting the overview table which will likely require us to first make some changes to the validations table.


tie-rack:
LGTM

PR #243, opened 2015-06-25
Refactor /db/ routes

tie-rack:
* Remove unused `getValidations`
* Use `feedid` url param consistently
* Update queries to use `results_id` now that that's standardized in data-processor
* Update `getValidationsErrorCount` to be a simple count on the validations table
* Sort routes in pg/services.js for niceness

Pivotal story: [97023298](https://www.pivotaltracker.com/story/show/97023298)


pg/gets.js
```diff
@@ -92,45 +92,14 @@ module.exports = {
...
-      for (i = 0; i < result.rows.length; i++) {
-        var parsed_message = edn.toJS(edn.parse(result.rows[i]["message"]));
-        result.rows[i]["message"] = parsed_message;
-      }
-
-      client.end();
-      resp.writeResponse(result.rows, res)
-    });
-  },
   getValidationsErrorCount: function(req, res) {
```
bwreid: This can probably just be changed to use the simpleQueryResponder, right?

tie-rack: Oh, you're right. I got sidetracked during this one. Fix coming.

bwreid:
Other than that, looks good to me.  :)

PR #244, opened 2015-06-26
[WR] connect single ballot page with pg

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/97582238)

Connects the single ballot page to the PostgreSQL database. Chris, would love your thoughts on the SQL queries in particular.


tie-rack:
:+1:

PR #253, opened 2015-07-06
Cleanup state overview pages

bwreid:
This fixes some of the links and tables on the Overview and State page/s.


tie-rack:
Solid. LGTM.

PR #254, opened 2015-07-07
Cleanup contests locality precincts and more

bwreid:
Fixes a number of pages to work with sortable data. Includes Contests, Locality, Precinct, Precinct Split, and Electoral District.


tie-rack:
LGTM. Let's do it!

PR #256, opened 2015-07-10
Error translations

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/98679906)

This adds more human readable text for our errors coming from Postgres.


tie-rack:
Awesome!

PR #257, opened 2015-07-13
[WR] complete hanging bracket; add error for invalid XML

bwreid:
Adds a translated error for XML Generation and fixes a hanging bracket on the errors page.


tie-rack:
Yep! :+1:

PR #259, opened 2015-07-17
Stormpath demo

tie-rack:
A preview of what integrating [Stormpath](https://stormpath.com/) could look like, and how we might use Stormpath's groups to facilitate authorization.

This PR can authenticate users against Stormpath, and changes the `/feeds` list page to only show feeds for a user's state (using the FIPS code in the source `vip_id`).

## Setup

Using Stormpath requires the following environment variables set:
```
STORMPATH_API_KEY_ID
STORMPATH_API_KEY_SECRET
STORMPATH_APP_HREF
STORMPATH_ACCOUNT_STORE
```

If they are all set, Metis will use Stormpath for authentication.

## Highlights

`authentication/utils.js` has some functions to deal with Stormpath groups and testing if a user is a "super admin".

The changes in `pg/gets.js` and `pg/queries.js` show how route handlers and SQL queries might need to change to include authorization checks. There is some room for improvement there, but this proves the concept.

## Where to go from here

* Pages beyond the `/feeds` page would need to have an authorization check that the user "owns" the feed. That should be possible with some sort of middleware I'm sure.
* Interfaces for user management (create, edit, delete, assign/remove from states).


bwreid:
I tried playing with this some. It seems like the better move is to use `express-stormpath` as it does many of the things we want in the future (eg. forgot password, creating new users) however it seems like a number of changes would need to take place with the routing in order to get this all to work properly.

Thankfully, we (1) would like to change the overall routing and (2) have access to [Stormpath's Angular Docs](http://docs.stormpath.com/angularjs/guide/index.html) which seem to be pretty detailed.

For now, I feel as though we've researched this option enough to know that it is able to provide a number of the features we need now (essentially already implemented with this research) and will want later (capable via express-stormpath or custom built) at a very reasonable price threshold.

The final issue of mapping between Groups and Feeds is a difficult one, but there is likely something clever we can do by looking up FIPS codes.

PR #260, opened 2015-07-20
[WR] show street segment errors for a specific locality

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/99182400)

Fixes a bug where street segment errors for a specific locality were not showing.

@tie-rack, is this inline with the way you set these up?


tie-rack:
Yep. That's the pattern.

LGTM.

PR #261, opened 2015-07-20
[WR] fix overview table and precincts table to show counts correctly

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/98682382)

Attempting to do completion calculations inside of the sql query was a bit too ambitious. This moves that computation out to the controller which can easily manage it and return the correct counts.


tie-rack:
LGTM.... thank you!!!!

PR #262, opened 2015-07-21
[WR] simple fix to change breadcrumb for election

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/98682714)

Super simple way to fix the above mentioned bug.


tie-rack:
LGTM

PR #263, opened 2015-07-22
[WR] update locality overview error pages to work correctly

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/99182400)

This update should, for real, fix all the error pages on the overview page for localities. It includes adding multiple lines to `gets_errors.js` and what necessarily follows from them.


tie-rack:
There we go. LGTM

PR #264, opened 2015-07-22
Add error count to every page

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/98680462)

Instead of attempting a new call inside of each controller, a new aside controller nested inside the main view handles querying for the total number of errors. Also did a bit of updating to `paths.js` to more adequately reflect the reality of the code base now.


tie-rack:
More deletes than adds. LGTM!

PR #266, opened 2015-08-04
[WR] updates query to correctly return counts for polling locations

bwreid:
Also updated election administration counts which were also wrong.

[Pivotal Card](https://www.pivotaltracker.com/story/show/98682382). This should fix the polling location counts on an individual locality page; but please give it a try yourself.


tie-rack:
:8ball: all signs point to yes

PR #268, opened 2015-08-17
Stormpath

tie-rack:
Add [Stormpath](https://stormpath.com/) integration for authentication and authorization.

This PR can authenticate users against Stormpath, and changes the `/feeds` list page to only show feeds for a user's state (using the FIPS code in the source `vip_id`).

## Setup

Using Stormpath requires the following environment variables set:
```
STORMPATH_API_KEY_ID
STORMPATH_API_KEY_SECRET
STORMPATH_APP_HREF
STORMPATH_ACCOUNT_STORE
```

If they are all set, Metis will use Stormpath for authentication.

This supersedes #259.


tie-rack:
All commits have been previously approved. Merging.

PR #269, opened 2015-08-18
Fix front end tests

bwreid:
Covers a number of cards. Primarily [this card](https://www.pivotaltracker.com/story/show/97112720) which is about fixing the front-end tests and is mostly solved with the first three commits. The following commits solve missing pages in order to eventually fix failing tests:

[Pivotal Card I](https://www.pivotaltracker.com/story/show/101219878)
[Pivotal Card 2](https://www.pivotaltracker.com/story/show/101215438)
[Pivotal Card 3](https://www.pivotaltracker.com/story/show/101214658)
[Pivotal Card 4](https://www.pivotaltracker.com/story/show/101211676)


pg-setup/database-setup.sql
```diff
@@ -0,0 +1,1790 @@
...
+--
+
+CREATE TABLE ballot_candidates (
+    results_id bigint NOT NULL,
+    ballot_id bigint,
+    candidate_id bigint
+);
+
+
+ALTER TABLE ballot_candidates OWNER TO wesreid;
```
tie-rack: `wesreid`?

bwreid: Yeah, that's not right. I can remove those comment lines but what to do about `OWNER TO`? It looks as though they're not needed to correctly run the import so maybe just remove?

tie-rack: Yeah. If it works without `OWNER TO`, getting rid of them is probably cool.

tie-rack:
Great. LGTM. This will come in handy!

PR #275, opened 2015-08-26
Redirect non-https clients to https

tie-rack:
AWS load balancers add an X-Forwarded-Proto header with the protocol used by clients. We want to redirect anyone coming over http to https.

A new /ping route has been added so that the load balancer has a page to health check on that clients won't care about. (The health check requests may have the X-Forwarded-Proto header set to "http" so it needs to be able to be accessed without an https redirect.)

Also, removed the never-used `https.createServer` call. The load balancer handles the SSL piece, so there's no need to do it in here now.

### DIFF
--- a/app.js
+++ b/app.js
@@ -6,7 +6,6 @@ var logger = (require('./logging/vip-winston')).Logger;
 var config = require('./config');
 var express = require('express');
 var http = require('http');
-var https = require('https');
 var path = require('path');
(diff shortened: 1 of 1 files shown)


bwreid:
LGTM

tie-rack:
Before merging, we need to update the load balancer's health check url to /ping

PR #277, opened 2015-08-31
Notifications

bwreid:
Tackles the following Pivotal Cards, with some slight changes:
[Pivotal Card: Send to Stormpath Groups](https://www.pivotaltracker.com/story/show/98952286)
[Pivotal Card: SES Research](https://www.pivotaltracker.com/story/show/101412072)
[Pivotal Card: Notification Works?](https://www.pivotaltracker.com/story/show/101124460)

This PR actually skips over using notification works and instead just sends an email directly form Metis. Until it becomes apparent we'll be sending many more emails, we think this is the most direct way to solve the issue. SES Research information will be added on to the Pivotal Card itself.

[Related work on a PR over in data-suite.](https://github.com/votinginfoproject/data-suite/pull/3)

### DIFF
--- a/notifications/content.js
+++ b/notifications/content.js
@@ -1,5 +1,9 @@
 module.exports = {
   processedFeed: function(message) {
     return 'Great job! You processed a feed. It has a public id. It\'s: ' + message[":public-id"] + '. Wow. Real cool.';
+  },
+  errorDuringProcessing: function(message) {
+    return 'It looks like a feed failed during processing. Here\'s the information we got: \
+            \nMessage we got: ' + JSON.stringify(message);
   }
 }
--- a/notifications/content.js
+++ b/notifications/content.js
@@ -1,0 +1,5 @@
+module.exports = {
+  processedFeed: function(message) {
+    return message;
+  }
+}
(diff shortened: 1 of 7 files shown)


notifications/content.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  processedFeed: function(message) {
+    return 'Great job! You processed a feed. It has a public id. It\'s: ' + message[":public-id"] + '. Wow. Real cool.';
```
tie-rack: Of course, we need real email content here.

Should we get it before merging this in? Or, since this will only send emails to me (for the moment, as configured), should we merge it now and make updating final copy a new story?

notifications/sender.js
```diff
@@ -0,0 +1,85 @@
...
+      to: recipient,
+      subject: 'Something Went Wrong with a Feed',
+      text: messageContent.errorDuringProcessing(rabbitMessage)
+    };
+  }
+};
+
+var sendMessage = function(messageContent) {
+  transporter.sendMail(messageContent, function(error, info) {
+    logger.info('Sending error: ' + error) if error;
```
tie-rack: You can't do that in Javascript! It's gotta be:

```js
if (error) logger.info('Sending error: ' + error);
```

tie-rack:
LGTM!

PR #278, opened 2015-08-31
[WR] remove lingering left joins

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/100812658)

Some remaining left joins were causing an empty row to come through. Changed both to inner joins and checked both pages.


tie-rack:
Oh great. LGTM.

PR #280, opened 2015-09-01
Only one email per processing complete message

tie-rack:
Share a queue bound to the "processing.complete" topic among all instances so that only one instance takes the message and sends emails about it. Otherwise, we send as many copies of the email as there are instances!

[Pivotal story](https://www.pivotaltracker.com/story/show/102519538)

### DIFF
--- a/notifications/queue.js
+++ b/notifications/queue.js
@@ -28,7 +28,7 @@ var onChannelOpen = function(err, ch) {
   var processingComplete = config.notifications.topics.processingComplete;
 
   ch.assertExchange(ex, "topic", exOptions);
-  ch.assertQueue('metisProcessingCompleteEmail', {}, function(err, ok) {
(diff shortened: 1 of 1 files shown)


notifications/queue.js
```diff
@@ -28,13 +28,14 @@ var onChannelOpen = function(err, ch) {
   var processingComplete = config.notifications.topics.processingComplete;
 
   ch.assertExchange(ex, "topic", exOptions);
-  ch.assertQueue('', {exclusive: true}, function(err, ok) {
+  ch.assertQueue('metisProcessingCompleteEmail', {}, function(err, ok) {
```
bwreid: We probably don't want to call it metis. Also, is camel case standard when we're talking about rabbit queues?

tie-rack: Looks like the standard is all lowercase with dots and underscores. I'll update.

bwreid:
Great. LGTM.

PR #281, opened 2015-09-10
[WR] add filter to properly add commas for addresses

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/100052082). Adds a comma after text elements if they exist. This allows us to dynamically and correctly format addresses if they are partially or wholly incomplete.

Although the card is about a single instance, I tried finding wherever this may show up.


tie-rack:
Smart! LGTM!

PR #283, opened 2015-09-17
include new relic on the dashboard front-end

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/102022928)

Takes environment variables to configure newrelic.js.


tie-rack:
So compact! LGTM!

PR #284, opened 2015-09-24
[WR] add ng-cloak to alert elements

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/102494524)

I believe this should solve the issue with flickering. [More information here.](https://docs.angularjs.org/api/ng/directive/ngCloak)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -139,11 +139,11 @@
           </ul>
         </nav>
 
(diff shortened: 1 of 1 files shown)


tie-rack:
Oh great. LGTM!

PR #285, opened 2015-09-24
[WR] do not throw error if rabbitmq unavailable

bwreid:
The logger error should be one of "failure," which should differentiate it from other errors.

[Pivotal Card](https://www.pivotaltracker.com/story/show/103420758)

### DIFF
--- a/notifications/queue.js
+++ b/notifications/queue.js
@@ -55,8 +55,8 @@ var url = 'amqp:' + config.notifications.host + ":" + config.notifications.port;
 var connect = function() {
   attempt += 1;
   if (attempt > 5) {
-    logger.error("Failed to connect to RabbitMQ!");
-    throw "RabbitMQ not available"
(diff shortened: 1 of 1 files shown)


notifications/queue.js
```diff
@@ -55,8 +55,8 @@ var url = 'amqp:' + config.notifications.host + ":" + config.notifications.port;
 var connect = function() {
   attempt += 1;
   if (attempt > 5) {
-    logger.error("Failed to connect to RabbitMQ!");
-    throw "RabbitMQ not available"
+    logger.log("failure", "Failed to connect to RabbitMQ!");
```
tie-rack: I wonder if, since we'll be sending these logs to syslog, the level should be one of the [standard syslog levels](https://en.wikipedia.org/wiki/Syslog#Severity_level). Maybe "crit" or "alert"?

bwreid: I'm into that! I'll change it.

PR #286, opened 2015-09-28
Loggly

tie-rack:
Simplify logging, while adding support for logging to Loggly via a linked Docker container.

* Remove complicated, unused file-based logging
* Remove unused winston-papertrail dependency
* Add winston-syslog dependency
* Configure and use syslog logging

Pivotal story: [102023084](https://www.pivotaltracker.com/story/show/102023084)


bwreid:
Looks Good To Me. :+1:

PR #287, opened 2015-10-01
update feed processed email

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/103455704)

Updates text for the processed feed email.

### DIFF
--- a/notifications/content.js
+++ b/notifications/content.js
@@ -1,6 +1,11 @@
 module.exports = {
-  processedFeed: function(message) {
-    return 'Great job! You processed a feed. It has a public id. It\'s: ' + message[":public-id"] + '. Wow. Real cool.';
+  processedFeed: function(message, recipient, group) {
+    return "<p>" + recipient.givenName + ",</p>" +
+           "<p>The data you provided for " + group.description + "'s election is available for you to review on the VIP Data Dashboard.</p>" +
+           "<p>Please click the link below to review your feed.</p>" +
+           "<p><a href='http://dashboard.votinginfoproject.org/#/feeds/" + message[":public-id"] + "'>Go to the Data Dashboard</a></p>" + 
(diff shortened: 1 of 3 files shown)


notifications/content.js
```diff
@@ -1,6 +1,11 @@
 module.exports = {
-  processedFeed: function(message) {
-    return 'Great job! You processed a feed. It has a public id. It\'s: ' + message[":public-id"] + '. Wow. Real cool.';
+  processedFeed: function(message, recipient, group) {
+    return "<p>" + recipient.givenName + ",</p>" +
+           "<p>The data you provided for " + group.description + "'s election is available for you to review on the VIP Data Dashboard.</p>" +
+           "<p>Please click the link below to review your feed.</p>" +
+           "<p><a href='http://dashboard.votinginfoproject.org/#/feeds/" + message[":public-id"] + "'>Go to the Data Dashboard</a></p>" +
```
tie-rack: How about the domain come from an environment variable (set in config.js maybe)? And then we could add it to consul so the these links would be right for both production and staging!

notifications/content.js
```diff
@@ -1,6 +1,13 @@
+var baseUrl = process.env.BASE_URI || "localdocker:4000";
+
 module.exports = {
-  processedFeed: function(message) {
-    return 'Great job! You processed a feed. It has a public id. It\'s: ' + message[":public-id"] + '. Wow. Real cool.';
+  processedFeed: function(message, recipient, group) {
+    return "<p>" + recipient.givenName + ",</p>" +
+           "<p>The data you provided for " + group.description + "'s election is available for you to review on the VIP Data Dashboard.</p>" +
+           "<p>Please click the link below to review your feed.</p>" +
+           "<p><a href='http://" + baseUrl + "/#/feeds/" + message[":public-id"] + "'>Go to the Data Dashboard</a></p>" +
```
tie-rack: This should be https.

tie-rack:
Cool. LGTM.

PR #288, opened 2015-10-02
Toggle approve feed button

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/102495008)

Shows a red button if the feed cannot be approved and a green button if it can.

### DIFF
--- a/notifications/sender.js
+++ b/notifications/sender.js
@@ -6,8 +6,12 @@ var messageContent = require('./content');
 var stormpathREST = require('stormpath');
 var pg = require('pg');
 
-var stormpathRESTApiKey = new stormpathREST.ApiKey(config.auth.apiKey, config.auth.apiKeySecret);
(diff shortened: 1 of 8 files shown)


notifications/sender.js
```diff
@@ -6,8 +6,12 @@ var messageContent = require('./content');
 var stormpathREST = require('stormpath');
 var pg = require('pg');
 
-var stormpathRESTApiKey = new stormpathREST.ApiKey(config.auth.apiKey, config.auth.apiKeySecret);
-var stormpathRESTClient = new stormpathREST.Client({ apiKey: stormpathRESTApiKey });
+if(config.auth.apiKey && config.auth.apiKeySecret) {
```
tie-rack: There's a function, `config.auth.uselocalauth()`, you can use here (and below) to test if Stormpath is configured.

tie-rack:
LGTM

PR #293, opened 2015-10-12
Detailed error box

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/103758704)

Middle two commits are the most crucial. This separates out the warnings from all other errors:

image


tie-rack:
I can accept that query... LGTM.

PR #294, opened 2015-10-13
[WR] remove errorHeadings in favor of detailing all error titles

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/101985558)

A complement to an already merged card from data-processor, this PR removes the slightly confusing 'errorHeadings' var and instead details out the full title for every error message. There is duplication, but it allows us to make more specific changes to individual titles for the future and solves the current problem where an example shows in both the title and the example.

image


tie-rack:
Makes sense. LGTM.

PR #295, opened 2015-10-20
Use connection pool facility of node-postgres

tie-rack:
Instead of creating a connection to postgres for each and every query, use [the connection pooling facility of node-postgres](https://github.com/brianc/node-postgres#client-pooling).

Fortunately, there are only three places where the bad style was being used.

Callback city.

Pivotal story: [103324962](https://www.pivotaltracker.com/story/show/103324962)


tie-rack:
I think they should definitely work for now. This is a very, very lightly used site atm.

PR #296, opened 2015-10-21
Fix sidebar data

tie-rack:
* Calculate feed due date from the election date for the feed
* Take title from breadcrumbs (it's how everywhere else has that title)
* Remove feed contact

Additionally (from discussions with Maria during the process):

* Only show feed title in the sidebar in "mobile"
* Make the two different feed "due date" calculations match up

Pivotal story: [105257072](https://www.pivotaltracker.com/story/show/105257072)

### DIFF
--- a/public/assets/js/app/app.js
+++ b/public/assets/js/app/app.js
@@ -293,7 +293,7 @@ vipApp.run(function ($rootScope, $appService, $location, $httpBackend, $appPrope
     $rootScope.getDueDateText = function(date){
 
       if(date){
-        var dueDate = moment(date, "YYYY-MM-DD").subtract($rootScope.$appProperties.electionDueDateWeeksInAdvance, 'weeks');
+        var dueDate = moment(date, "YYYY-MM-DD").subtract(23, 'days');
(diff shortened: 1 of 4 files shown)


public/assets/js/app/app.js
```diff
@@ -293,7 +293,7 @@ vipApp.run(function ($rootScope, $appService, $location, $httpBackend, $appPrope
     $rootScope.getDueDateText = function(date){
 
       if(date){
-        var dueDate = moment(date, "YYYY-MM-DD").subtract($rootScope.$appProperties.electionDueDateWeeksInAdvance, 'weeks');
+        var dueDate = moment(date, "YYYY-MM-DD").subtract(23, 'days');
```
ericnormand: Just curious here: hardcoded 23 days is ok? It's not a number of weeks.

tie-rack: Yeah, it's no longer a number of weeks. It's to get it to match [this date calculation](https://github.com/votinginfoproject/Metis/blob/master/public/assets/js/app/controllers/feedsController.js#L19), which uses 22 instead because everyone does date math on the boundary different.

ericnormand:
LGTM

PR #297, opened 2015-10-23
Get all dependency downloading cached

tie-rack:
The "onbuild" version of the node docker image was forcing us to download some dependencies on every build, since it happened after copying code over.

This moves to the unadorned node base image, and handles all dependency downloading before the `COPY . /usr/src/app` directive, so they will be cached for the vast majority of new builds.


ericnormand:
Nice. It builds fast the second time really fast.

LGTM

PR #298, opened 2015-10-26
Correct acceptable electoral district values

tie-rack:
The old list in the error message does not match the actual validation. The error message has been updated with the actual regular expression used.

Pivotal bug: [105787956](https://www.pivotaltracker.com/story/show/105787956)


ericnormand:
Do we show error messages with regexes?

LGTM

tie-rack:
@ericnormand we do now!

PR #299, opened 2015-10-26
Fix 500s when trying to download the error report

tie-rack:
When we updated how we use node-postgres to use connection pooling (#295), we forgot to update the query that builds the CSV error report, causing an internal server error when trying to download one. This updates the query so it can work again.

There are no other remnants of the old `openPostgres` function (verified with grep).

Pivotal bug: [106671418](https://www.pivotaltracker.com/story/show/106671418)


ericnormand:
LGTM

PR #301, opened 2015-10-29
Confirm approval

tie-rack:
Add a popup confirmation when a user hits the "approve feed" button. Uses the text from the pivotal story.

Additionally, update a few changes that got left behind on Wesley's laptop:
* Log contents of emails that fail to send
* Correct email body
* Use the correct callback for sending feed approval emails on the "approve feed" button

Pivotal story: [105631088](https://www.pivotaltracker.com/story/show/105631088)


ericnormand:
LGTM

PR #302, opened 2015-10-30
Handle errors in simpleQueryResponder

tie-rack:
Instead of blowing up because result is null, return a 500, log it, and let the server not die.


ericnormand:
LGTM

PR #303, opened 2015-11-03
Change approve button on approved

tie-rack:
When a user clicks the "Approve feed" button, the database is updated and the button turns into the darker "Approved" non-clickable button.

Pivotal story: [106138734](https://www.pivotaltracker.com/story/show/106138734)


dcohen:
LGTM - the only thing I think I'd like to see is a less ugly hover color, but everything else is fine.

PR #304, opened 2015-12-03
Add time to "Loaded on" column of feeds list

tie-rack:
Query updated to no longer drop the time portion of the `r.start_time`.

Pivotal story: [107533056](https://www.pivotaltracker.com/story/show/107533056)


tie-rack:
feed-list-with-time

ericnormand:
LGTM

PR #305, opened 2015-12-14
Update queries that rely on 3.0 tables

tie-rack:
`results`, `validations`, and `election_approvals` are not version-specific, so they don't have a version prefix and are not renamed.

Pivotal story: [109360032](https://www.pivotaltracker.com/story/show/109360032)


ericnormand:
LGTM

PR #306, opened 2015-12-18
Namespace v3 endpoints

tie-rack:
Routes that hit "database endpoints" are new broken out if they are specific to a specification version.

Pivotal story: [109361178](https://www.pivotaltracker.com/story/show/109361178)


dcohen:
LGTM!

PR #310, opened 2016-04-06
Fix polling location count on locality page

tie-rack:
Gotta join on the right thing. We had been joining `v3_0_precinct_split_polling_locations` trying to match its `precinct_split_id` to a *precinct's* id, not to a *precinct split's* id. Oops.

Pivotal story: [116323479](https://www.pivotaltracker.com/story/show/116323479)


cap10morgan:
LGTM

PR #313, opened 2016-05-02
Paginate feeds in chunks of 20

sduckett:
Rendering the dashboard was slow, so we decided to paginate the results. 20 seems like a good number for now.

[Pivotal card](https://www.pivotaltracker.com/story/show/116163065)


7maples:
LGTM!!!

PR #314, opened 2016-05-05
Send email when xml feed completes

sduckett:
The 3.0 and 5.0 feeds have results in different places, so in this PR, we take the message announcing a successful run and looks in both places to find the result before sending an email announcing the details of the run.

This is a step that's required to be finished before we can merge data-processor's `import-5.0-xml-into-postgres` branch.

[Pivotal card](https://www.pivotaltracker.com/story/show/116162051)

### DIFF
--- a/notifications/content.js
+++ b/notifications/content.js
@@ -15,6 +15,14 @@ module.exports = {
            "<p>If you have any questions, please contact <a href='mailto:vip@democracy.works'>vip@democracy.works</a>.</p>" +
            "<p>Thank you!</p>";
   },
+  v5processedFeed: function(message, recipient, group) {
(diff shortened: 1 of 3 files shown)


cap10morgan:
Gross that the log messages had to change to `info` level. But I assume you spent a reasonable amount of time investigating and couldn't find a solution. So unless @tie-rack knows what the problem might be, this LGTM.

tie-rack:
We spent the time and just don't know. So `info` it is.

cap10morgan:
In case you were still waiting on me, my previous comment was intended to boil down to "LGTM"

tie-rack:
Not waiting on you :)

This is waiting for a coordinated deploy of data-processor.

PR #316, opened 2016-05-17
Restart service on failure

tie-rack:
`--restart=always` in the ExecStart command doesn't work, and interferes with the `Restart=on-failure` service directive.

Only took a day and a half to get to this point.


sduckett:
LGTM, @tie-rack.

PR #320, opened 2016-07-05
5.1 overview starts with a title

sduckett:
If the `public_id` is built correctly, we can format it and get a nice title for our overview page.
[Pivotal card](https://www.pivotaltracker.com/story/show/123237561)


pg/v5_1_queries.js
```diff
@@ -0,0 +1,12 @@
+var util = require('./util.js');
+var overviewQuery = "select r.id, state.name as state_name, \
+                     elections.election_type as election_type, \
+                     to_char(to_date(elections.date, 'MM/DD/YYYY'), 'YYYY-MM-DD') as election_date \
+                     from results r \
+                     left join v5_1_states state on state.results_id = r.id \
+                     left join v5_1_elections elections on elections.results_id = r.id \
```
tie-rack: These `v5_1_*` tables won't be populated for XML feeds.

public/assets/js/app/app.js
```diff
@@ -795,4 +795,20 @@ vipApp.run(function ($rootScope, $appService, $location, $httpBackend, $appPrope
...
+  /*
+   * TODO figure out where this stuff _ACTUALLY_ goes...
+   * TODO perhaps lose these helpers and find a better way to do the thing...
+   */
+  $rootScope.v5_utils = {};
+  $rootScope.v5_utils.feedId = function () {
+    var locationParts = window.location.hash.split("/");
+    if (locationParts[2] === "feeds") {
+      return locationParts[3];
+    }
```
tie-rack: Should there be an `else` clause here? What happens if `locationParts[2]` isn't "feeds"?

public/assets/js/app/controllers/5.1/feedOverview51Controller.js
```diff
@@ -1,7 +1,19 @@
 'use strict';
 
-function FeedOverview51Ctrl($scope, $routeParams) {
-  var publicId = $routeParams.vipfeed;
-  $scope.publicId = publicId;
-  $scope.errorReport = "/db/feeds/" + publicId + "/xml/errors/report";
+function FeedOverview51Ctrl($scope, $rootScope, $feedDataPaths, $routeParams, $location, $appProperties, $filter, ngTableParams) {
```
tie-rack: Are any of `$location`, `$appProperties`, `$filter`, or `ngTableParams` used here? It doesn't look like it.

public/assets/js/app/controllers/5.1/v5AsideController.js
```diff
@@ -0,0 +1,3 @@
+function v5AsideController($scope, $rootScope, $http, $feedDataPaths, $routeParams) {
```
tie-rack: Only `$routeParams` seems to be used here, do the others need to be passed in?

public/assets/js/app/app.js
sduckett: You'd get a `null`, which would route back to the feed list. I'm punting on this part right now; let's talk about it tomorrow, @tie-rack.

PR #321, opened 2016-07-11
Include error counts in 5.1 feed overviews

sduckett:
This PR adds high level info about errors in your feed. **Total Errors in Feed** are a combination of errors with severities of `fatal`, `critical`, and `errors`; **Warnings in Feed** are just the `warnings`. See the [Pivotal card](https://www.pivotaltracker.com/story/show/123263841) for more infomation.

You might look at `index.html` and say to yourself: "*Why would we add an `ng-if` directive? We don't need it here!*" If the world were a sane place, you'd be right, but the world is not always so. It turns out that if you omit this directive, `V5AsideController` will be called *before* `FeedOverview51Ctrl`, and `$routeParams.vipfeed` will be `undefined`. You won't be able to request any errors with an undefined `public_id`! 

"*There must be a better way!*", I can hear you saying, and I hope you're right. Let's talk about that part and see if we can find the better way.

### DIFF
--- a/pg/services.js
+++ b/pg/services.js
@@ -50,6 +50,7 @@ function registerPostgresServices (app) {
   ///////// Version 5.1 /////////
   app.get('/db/feeds/:feedid/xml/overview', pg51.feedOverview);
   app.get('/db/feeds/:feedid/xml/errors/report', csv.xmlTreeValidationErrorReport);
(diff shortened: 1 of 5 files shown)


tie-rack:
I'm afraid "a better way" would involve starting from scratch, so this LGTM.

PR #323, opened 2016-08-01
5.1 errors page

tie-rack:
An errors overview page for 5.1 feeds. Splits fatal and critical errors into a table above other errors.

Pivotal story: [123274071](https://www.pivotaltracker.com/story/show/123274071)


sduckett:
LGTM

PR #324, opened 2016-08-08
5.1 overview tables

tie-rack:
Adds the "Voter Resources", "Contests", and "Source and Election" tables to the overview page for 5.1 feeds.

Along the way, I created an Angular directive `<vip-overview-table>` to make things easier. It turns out there was an old directive that was supposed to do this, but it wasn't used anywhere and didn't work. I removed that one.

Pivotal stories: [123274731](https://www.pivotaltracker.com/story/show/123274731), [123274501](https://www.pivotaltracker.com/story/show/123274501), [123274801](https://www.pivotaltracker.com/story/show/123274801)


public/app/partials/5.1/feed-overview.html
```diff
@@ -1,91 +1,36 @@
...
+       table-params="tableParams.contests"
+       title="'Contests'"
+       id-prefix="'contest'">
+    </vip-overview-table>
+
+    <vip-overview-table
+       table-data="summaries.sourceElection"
+       table-params="tableParams.sourceElection"
+       title="'Source and Election'"
+       id-prefix="'sourceElection'">
```
sduckett: Could you remind me why we need to write `"'sourceElection'"` here?

tie-rack: Without the subquotes, it's treated as a property to look up on the scope. With the subquotes, it's a string literal.

public/app/partials/5.1/feed-overview.html
```diff
@@ -1,91 +1,36 @@
 <i id="feed-overview-content"></i>
 
 <div class="overview-wrap">
-  <section class="data-group data-module started">
-    <header class="data-group-header">
-      <h2>Voting Locations</h2>
-      <span ng-if="!overviewData" class="more-detail is-loading"></span>
+  <section class="data-group data-module started"
+           ng-show="summaries">
+    <vip-overview-table
```
sduckett: Directives make this much more readable 👍

public/app/partials/5.1/feed-overview.html
tie-rack: Though... it looks like we could use `'@'` instead of `'='` in the directive for those bindings and we won't need the extra quotes.

sduckett:
LGTM

PR #325, opened 2016-08-17
Increase postgres timeouts

sduckett:
Some states have feeds with millions of errors in their street segment
files. The queries and CSV generation take more time that we were
allowing, and when the timeout hit, we'd return them a file with only
header rows.

Three minutes seems like a lot, but it allows the queries to finish. On my dev machine, runtime is ≈ 2.5 minutes; on the staging nodes it is ≈ 1 minute.

### DIFF
--- a/pg/conn.js
+++ b/pg/conn.js
@@ -9,7 +9,7 @@ var config = {
   password: process.env.DB_ENV_POSTGRES_PASSWORD,
   port: process.env.DB_PORT_5432_TCP_PORT,
   max: 10, // clients in the connection pool
-  idleTimeoutMillis: 500000
+  idleTimeoutMillis: 300000
 };
 
 var pool = new pg.Pool(config);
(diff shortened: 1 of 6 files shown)


pg/conn.js
```diff
@@ -2,15 +2,35 @@ var resp = require('./response.js');
...
+  max: 10, // clients in the connection pool
+  idleTimeoutMillis: 300000
+};
+
+var pool = new pg.Pool(config);
+
+var queryFromPool = function(callback) {
+  pool.connect(function(err, client, done) {
+    if(err) {
+      return console.error('error fetching client from pool', err);
```
tie-rack: This used to be `logger.crit`. Why the change? And do messages sent to `console` end up using the logging configuration so that messages in production end up in our log aggregator? I'm guessing not.

sduckett: No, that shouldn't have changed.

tie-rack: So what about the addition of `console.error` to `logger.error` below? Do we need both, or would `logger.error` be enough?

sduckett: > Do we need both, or would `logger.error` be enough?

I'd like it to be enough, but didn't we run into some problem with that level not making it to the logs?

tie-rack: Hmm... maybe. But `console.error` is definitely not going to make it to Loggly.

sduckett:
I've confirmed that Loggly is actually picking up `error` severity now. No more writing `logger.info("[ERROR] mess");`

tie-rack:
Heck yes!

PR #326, opened 2016-08-17
Count the actual number of polling locations

sduckett:
Those inner join operations yield four duplicate rows. We found it by
noticing one locality had more polling locations than the entire state.


tie-rack:
LGTM 🚴

PR #327, opened 2016-08-24
Voting locations error page

sduckett:
This PR lets you click a link next to an element in the overview's "Voting Locations" table and see just that portion of a feed's errors, e.g. only problems with Locality elements. There's a little cleanup of how we show error_type since `missing` seems a little curt; there are a few more as we encounter them.

[Pivotal card](https://www.pivotaltracker.com/story/show/126628295)

screen shot 2016-08-24 at 11 30 35 am


tie-rack:
LGTM! 🔒

PR #333, opened 2016-09-02
V5 localities overview

sduckett:
We add a paginated table that shows you the _Localities_ of your feed, very much like what you'd find on a v3 feed's overview. One difference is that the locality's `id` isn't being coerced into an integer; this is useful for those hand-generated feeds where IDs look are alphanumeric, e.g., `loc001` where such coercion would yield `NaN`. Another difference is that we're sorting by the name instead of the ID; it seems to be more humane and I like it.

You might notice use of `crosstab` in the query - this is from PostgreSQL's [tablefunc](https://www.postgresql.org/docs/9.4/static/tablefunc.html) extension. We'll need to `create extension tablefunc;` on the DB before this PR gets merged.

[Pivotal story](https://www.pivotaltracker.com/story/show/123274871)


tie-rack:
@sduckett create a migration in data-processor for creating the extension, and then let's see this on staging.

tie-rack:
LGTM

PR #334, opened 2016-09-06
Fix error link buttons

sduckett:
When there are no errors of a specific type, linking to the errors gives a blank page. This fixes that.

[Bug #1](https://www.pivotaltracker.com/story/show/128178757) and [bug #2](https://www.pivotaltracker.com/story/show/129631863).


tie-rack:
LGTM 📶

PR #335, opened 2016-09-13
Don't count the locality with the precincts

sduckett:
If we don't restrict out join with `AND precincts.simple_path = 'VipObject.Precinct.LocalityId'`, we'll  get one more than we should. 

https://www.youtube.com/watch?v=1fbMQ5Z2lN4


tie-rack:
LGTM 👡

PR #336, opened 2016-09-15
Little application tweaks

sduckett:
Adding the application name is useful in looking at where a query comes from. Compressing the response is a good idea.


tie-rack:
This looks appropriate and good.

PR #338, opened 2016-10-10
Remove patch level for urls

tie-rack:
A companion to votinginfoproject/data-processor#229. With patch levels included in the `spec_version` but 5.1 feeds all sharing a set of views, chop off the patch for the URL.

### DIFF
--- a/public/assets/js/app/app.js
+++ b/public/assets/js/app/app.js
@@ -777,6 +777,10 @@ vipApp.run(function ($rootScope, $appService, $location, $httpBackend, $appPrope
     }
   }
 
+  var versionWithoutPatch = function(version) {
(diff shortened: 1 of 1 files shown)


public/assets/js/app/app.js
```diff
@@ -777,6 +777,10 @@ vipApp.run(function ($rootScope, $appService, $location, $httpBackend, $appPrope
     }
   }
 
+  var versionWithoutPatch = function(version) {
+    return version.replace(/^(\d+\.\d+)\.\d+$/, "$1");
```
sduckett: If we ever get "5.1.2.a", this will break. However, if that ever happens, we'll have other problems.

PR #345, opened 2016-12-02
5 1 source and election

sduckett:
This adds a page that displays the interesting parts of source and election elements from a feed. Most of the interesting work is in this *blocking* [data-processor PR](https://github.com/votinginfoproject/data-processor/pull/241)

[Pivotal card](https://www.pivotaltracker.com/story/show/128176637)

screen shot 2016-12-02 at 3 13 08 pm

### DIFF
--- a/public/app/partials/5.1/aside.html
+++ b/public/app/partials/5.1/aside.html
@@ -20,7 +20,7 @@
     <nav class="feed-nav">
       <ul>
         <li><a id="sidebar-overview" href="{{feedURL('')}}">Overview</a></li>
-        <li><a id="sidebar-source" href="{{feedURL('/source')}}">Source & Election</a></li>
+        <li><a id="sidebar-source" href="{{feedURL('/source')}}">Source &amp; Election</a></li>
         <li><a id="sidebar-errors" href="{{feedURL('/errors')}}">Errors</a></li>
       </ul>
     </nav>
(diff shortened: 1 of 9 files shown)


public/app/partials/5.1/aside.html
```diff
@@ -20,7 +20,8 @@
     <nav class="feed-nav">
       <ul>
         <li><a id="sidebar-overview" class="is-finished" href="{{feedURL('')}}">Overview</a></li>
-        <li><a id="sidebar-source" class="is-finished" href="{{feedURL('/errors')}}">Errors</a></li>
+        <li><a id="sidebar-source" class="is-finished" href="{{feedURL('/source')}}">Source & Election</a></li>
```
tank157: Could consider html escaping the `&` here. Probably not a huge deal.

public/app/partials/5.1/source.html
```diff
@@ -0,0 +1,98 @@
+<p ng-if="source.error_count !== undefined">
```
tank157: Does this work if `source` is not yet defined? It seems like the code below in the loading and data module sections is checking at that level.

tank157: Similar question with election error count further down.

public/app/partials/5.1/aside.html
sduckett: Good idea!

public/app/partials/5.1/source.html
sduckett: [In the controller](https://github.com/votinginfoproject/Metis/pull/345/files#diff-db1a977b0c3b8acc636d22275df68040R11) we add these entities to our `$scope`.

PR #353, opened 2017-02-10
V 3 polling location count for locality

katrinamariehh:
Adding `polling location` counts to the 3.0 feed overview localities table for [this story](https://www.pivotaltracker.com/story/show/136985505).

### DIFF
--- a/pg/queries.js
+++ b/pg/queries.js
@@ -297,8 +297,11 @@ module.exports = {
                                  GROUP BY p.id, p.name;",
     localities: "with \
                   precincts as \
-                    (select results_id, id, locality_id \
-                     from v3_0_precincts), \
+                    (select results_id, v3_0_precincts.id, locality_id \
+                     from v3_0_precincts \
+                     left join results r \
+                      on v3_0_precincts.results_id = r.id \
+                     where r.public_id = $1), \
                   precinct_splits as \
                     (select \
                        pl.id, \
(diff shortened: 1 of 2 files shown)


pg/queries.js
```diff
@@ -295,14 +295,44 @@ module.exports = {
...
+                  count(distinct p.id) as precincts, \
+                  count(distinct pl.id) as polling_locations, \
+                  count(distinct pspl.id) as split_polling_locations \
+                from results r \
+                left join v3_0_localities l on l.results_id = r.id \
+                left join precincts p on p.locality_id = l.id and p.results_id = r.id \
+                left join polling_locations pl on pl.precinct_id = p.id \
+                left join precinct_splits pspl on pspl.precinct_id = p.id \
+                where r.public_id = $1 \
+                group by p.locality_id, l.name;",
```
tie-rack: There are a number of seq scans in the `EXPLAIN` output for this query. For example for the `select results_id, id, locality_id from v3_0_precincts` subquery. Adding constraints on the `results_id` should help.

katrinamariehh: If I can get rid of all of the seq scans, will that satisfy this ask?

tie-rack: Yes!

katrinamariehh:
@tie-rack I fixed one of the seq scans with the change I just committed; for the other one I ran `explain analyze` and while the explain estimates it with a really high cost, the actual time for the query is not that long (i.e. with 58 3.0 feeds in my database the query only takes 14.671 ms even though the cost estimate is `cost=10000000000.00..10000000025.10`).  I'm happy to keep at trying to figure out how to improve it, but I'm not sure how much time we want to spend on it if we can't see results for it

PR #356, opened 2017-04-04
Add vit staging tool page

tank157:
Adds the VIT staging landing pages for both 3.0 and 5.1 feeds, as per [this Pivotal Card](https://www.pivotaltracker.com/story/show/141707995).

Sorry about the slightly out-of-order merging in of master at the end rather than the start, but there were no overlaps so seemed innocuous.


tank157:
The location of the VIT is going to change per discussions.

PR #366, opened 2017-05-23
Upload file

katrinamariehh:
This PR covers the work in [these](https://www.pivotaltracker.com/story/show/145129299) [three](https://www.pivotaltracker.com/story/show/145068793) [stories](https://www.pivotaltracker.com/story/show/145129509).

## Endpoint for uploading files
In [this](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-19d94c7e716be186c5f14b5d9820eaac) file an endpoint is created at `/testing/upload` to accept `POST` requests for files to be uploaded; the endpoint is registered [here](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-0364f57fbff2fabbe941ed20c328ef1a) to make it available as part of the main node app.  The upload functionality itself is handled [here](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-23ff561311fe89590b7c66154055cd66) in the `posts.js` file where we upload a file to s3 using the `aws-sdk` function `putObject`.  The file retains the name it had upon being uploaded and is saved in a directory with the structure `address-testing/<users_group_fips_code>/input/`.

## The Node app needs to notify workers that a file has been uploaded
When the file upload defined above is successful, there is a [callback](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-23ff561311fe89590b7c66154055cd66R37) to a function in `data-testing/queue.js` [file](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-dd973e5c92fab03b0039da8035998ff1R14) which converts the bucket and filename to edn and submits the message to the `batch-address.file.submit` queue.  

We additionally did some refactoring around the code to setup and utilize the queues in the Metis project overall; previously all of the information for accessing and setting up queues was housed in `notifications/queue.js`; shared code about connecting lives in `queue.js` and code about the specific queues used in the project are now in `notifications/queue.js` and `data-testing/queue.js`.

## Upload the file to be tested to the backend Node app
The upload button on [this page](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-64c482d01057999d1f13929792678237) was set up to function properly, including a callback that was added to that page's controller [here](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-f1ce968bf89771f019cec51d88f98574).  `ng-file-upload` was added to the main VIP app [here](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-1b523fb2f039b6893e584db51c14a644) as well as being added to the main `index.html` [here](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-528c3923d718a8860f5d8c05c3931c55R50) and to the utilities installed by bower [here](https://github.com/votinginfoproject/Metis/compare/testing-tools...upload-file?expand=1#diff-0a08a7565aba4405282251491979bb6b).


data-testing/posts.js
```diff
@@ -0,0 +1,47 @@
+var AWS = require('aws-sdk');
```
katrinamariehh: I went with `posts.js` because the file with `GET` request definitions in it is called `gets.js` but I think you're probably right that this is maybe not the clearest.  Would expanding it out to something like `postRequests.js` make it clearer?

data-testing/services.js
```diff
@@ -0,0 +1,7 @@
+var post = require("./posts.js");
```
katrinamariehh: For the LOLs!  The `gets.js` file that I was basing some of this structure around (see other comment) is actually required as `pg` and I was 100% sure what comparatively this new file should be called.  Does changing the import to `var posts = require("./posts.js")` make it cleaner/clearer?

data-testing/posts.js
sduckett: What do you think? Does calling it `postRequests` clearly describe the things that this module / chunk does?

katrinamariehh: I think if we want to be really descriptive and it doesn't necessarily need to follow the naming conventions of `gets.js` that lead me to call it `posts.js` it could maybe just be called `requests.js` or something more generic; at some point we'll need to connect everything together to check for files in a user/group's `output` folder and with the configuration already in the file is might make sense for those things to go together and then in *that* case we're talking about a `GET` request instead of a `POST` request and then `posts.js` _really_ doesn't make sense.  So I'm thinking maybe something like `requests.js` or something like `handler.js` seems likely.  Thoughts? @tank157

PR #372, opened 2017-06-26
fixed fileName and bucketName to match conventions on batt

katrinamariehh:
There was a mismatch between here and the batch-address-test-tool in how we were naming files for s3 (i.e. `bucketName` = `address-testing/<groupName>/<input/output>` with `fileName` = `<some_file>.csv` vs. `bucketName` = `address-testing` with `fileName` = `<groupName>/<input/output>/<some_file>.csv`); the second one is cleaner for `batch-address-test-tool` (i.e. we can construct the name for the output file more easily in this form) so that's what I'm going with.  Batch-Address-Test-Tool PR [here](https://github.com/votinginfoproject/batch-address-test-tool/pull/6)

### DIFF
--- a/data-testing/aws.js
+++ b/data-testing/aws.js
@@ -30,8 +30,8 @@ module.exports = {
         if (groupName === undefined) {
           groupName = "undefined"
         };
-        var bucketName = 'address-testing/' + groupName + '/input';
-        var fileName = files.file.originalFilename;
+        var bucketName = 'address-testing';
+        var fileName = groupName + '/input/' + files.file.originalFilename;
         s3.putObject({
           Bucket: bucketName,
           Key: fileName,


data-testing/aws.js
```diff
@@ -30,8 +30,8 @@ module.exports = {
         if (groupName === undefined) {
           groupName = "undefined"
         };
-        var bucketName = 'address-testing/' + groupName + '/input';
-        var fileName = files.file.originalFilename;
+        var bucketName = 'address-testing';
```
tank157: Approving this, but as a future improvement, would be nice to get the bucketName from configuration so we can also transition them over time as we roll out new infrastructure.

katrinamariehh: Good call!  Let's make sure to make a card for that.

PR #376, opened 2017-07-19
Env specific batt buckets

katrinamariehh:
This PR puts the bucket name used by the Batch Address Test Tool (which is dictated by the bucket Metis uploads to; it uploads the file submitted by the user and sends the message to the BATT telling it where the file it should process is) into consul to make sure that staging and production aren't using the same buckets (they were before).


katrinamariehh:
Merged into #377.

PR #377, opened 2017-07-31
Swap Stormpath for Auth0

tank157:
Despite the branch name, this ended up being the long running branch for all the work to transition from Stormpath to Auth0.

Some highlights:
* Authentication was happening in the express app side with Stormpath, but now with Auth0 it happens in the Angular side primarily.
* The express app is passed an access token to get user data for the front end.
* We had to set up a non-interactive client configuration for the express app to search user data on Auth0 in response to file processing messages.
* Protected API calls check a JWT that is included in the headers.
* Configuration has to be generated for the Angular app from the environment, leading us to rely more on grunt now.
* We had to jump many versions of node to be compatible with the jwt auth libraries. Everything still works as far as we can tell.


Dockerfile
```diff
@@ -1,6 +1,10 @@
-FROM node:0.12.15
+FROM node:7.10.0
 MAINTAINER Democracy Works, Inc. <dev@democracy.works>
 
+RUN apt-get update && \
+    apt-get install -y ruby rubygems-integration inotify-tools && \
```
tank157: There's probably another way to get `sass` generation in node rather than installing ruby, but this does work. Note that `sass` compilation is now achieved via `grunt` tasks.

authentication/services.js
```diff
@@ -0,0 +1,100 @@
...
+    jwksUri: "https://" + config.auth0.domain + "/.well-known/jwks.json"
+  }),
+
+  // Validate the audience and the issuer.
+  aud: config.auth0.audience,
+  iss: "https://" + config.auth0.domain + "/",
+  algorithms: ['RS256']
+});
+
+function checkAuth(scope_array) {
```
tank157: We never ended up using this, but if we ever do implement custom scopes and scoped API access, it could come in handy.

authentication/services.js
```diff
@@ -0,0 +1,100 @@
...
+      console.log(err);
+    } else {
+      console.log("getUser success?!?");
+    }
+    cb(user);
+  });
+}
+
+function registerAuthServices(app) {
+  app.post('/services/getUser', function(req, res) {
```
tank157: Wondering if we should add the `checkJwt` middleware here too.

data-testing/sender.js
```diff
@@ -47,54 +40,42 @@ var sendMessage = function(messageContent) {
   });
 };
 
-var notifyGroup = function(message, groupName, contentFn) {
-  if ((typeof groupName != "string") ||
-       (groupName.length < 2) ||
-       (groupName.length > 5)) {
-    logger.info("groupName is bad--sending to admin group");
-    groupName = config.email.adminGroup;
+var notifyGroup = function(message, fips, contentFn) {
```
tank157: @katrinamariehh , think we should change the name here? Group was really a nod to the Stormpath Groups concept, but it still sort of fits as a collection of people to notify. Just wondering if maybe we should change it to a Fips/State kind of name.

data-testing/sender.js
```diff
@@ -47,54 +40,42 @@ var sendMessage = function(messageContent) {
...
   sendNotifications: function(message) {
-    if(config.auth.uselocalauth()) {
-      logger.warning('A message was trying to be sent but cannot be Stormpath \
-                      credentials are not set! Message: ' + message);
-    } else {
-      var messageType  = (message['status'] == "ok") ? messageOptions['testingComplete'] : messageOptions['errorDuringTesting'] ;
-      var groupName = message["groupName"];
-      if (groupName === undefined) {
-        logger.warning("No group in batch-address.file.complete message.  Can't send batch address testing finished email notification.");
+    // #TODO-auth need to confirm connected to authentication source
```
tank157: We can remove this comment.

data-testing/sender.js
```diff
@@ -47,54 +40,42 @@ var sendMessage = function(messageContent) {
...
+    var messageType  = (message['status'] == "ok") ? messageOptions['testingComplete'] : messageOptions['errorDuringTesting'] ;
+    var groupName = message["groupName"];
```
tank157: Here's `group` again, though to change this would require changes to the Batch Address Test Tool codebase as well.

notifications/sender.js
```diff
@@ -64,41 +57,27 @@ var sendMessage = function(messageContent) {
   });
 };
 
-var notifyGroup = function(message, groupName, contentFn) {
-  if ((typeof groupName != "string") ||
-       (groupName.length < 2) ||
-       (groupName.length > 5)) {
-    logger.info("groupName is bad--sending to admin group");
-    groupName = config.email.adminGroup;
+var notifyGroup = function(message, fips, contentFn) {
```
tank157: Another instance of `group`

data-testing/sender.js
katrinamariehh: I think for this one we should maybe make a card just to make sure to keep the two projects coordinated.

public/assets/js/app/controllers/feedsController.js
```diff
@@ -3,26 +3,30 @@
  * Feeds Controller
  *
  */
-function FeedsCtrl($scope, $rootScope, $feedDataPaths, $feedsService, $location, $filter, ngTableParams, $interval, $timeout, $route) {
-
+function FeedsCtrl($scope, $rootScope, $feedDataPaths, $feedsService, $location, $filter, ngTableParams, $interval, $timeout, $route, $authService) {
+  console.log("at top of FeedsCtrl");
```
tank157: Seems like a log statement we can pull.

data-testing/sender.js
katrinamariehh: Yes--maybe now that we're considering 5-digit FIPS codes keeping it in terms of FIPS codes would be a good move?

public/assets/js/app/controllers/feedsController.js
```diff
@@ -3,26 +3,30 @@
...
-       key: "feeds",
-       errorMessage: "Could not retrieve Feeds."},
-      function(result) {
-        for (var i = 0; i < result.length; i++) {
-          var feed = $rootScope.feeds[i];
-          var date = new Date(feed.election_date);
-          $rootScope.feeds[i]['due_on'] = date.setDate(date.getDate() - 22);
-        }});
+    $authService.getUser(function (user){
+      console.log(user);
```
tank157: Pull this log statement, or change it to not expose all the user data maybe?

katrinamariehh: Ooh, yeah, we don't want to expose all of that data in the console I don't think.

public/assets/js/app/services/appService.js
```diff
@@ -5,13 +5,22 @@
...
+
 
     return {
-        // gets the User object from the server
-        getUser: function () {
-            // adding in the timestamp so that IE does not Cache this service
-            return $http.get($appProperties.servicesPath + "/getUser" + vipApp_ns.cacheBuster() );
+        // gets the User object from the server and updates root scope
+        setUserSuccess: function (data) {
+          console.log("at top of setUserSuccess");
```
tank157: change/remove log statement

public/assets/js/app/controllers/feedsController.js
katrinamariehh: 👍🏻

Dockerfile
```diff
@@ -18,8 +22,7 @@ RUN bower --allow-root install
 
 COPY . /usr/src/app
 
-RUN node-sass public/assets/css/app.scss public/assets/css/app.css
-
 EXPOSE 4000 27017 28017
 
-CMD [ "npm", "start" ]
+ENTRYPOINT [ "grunt" ]
+CMD [ "grunt",  "default"]
```
tie-rack: What's in `CMD` gets passed as args to the `ENTRYPOINT`, so this will end up trying to run `grunt grunt default` and it will complain that the `grunt` task isn't found. Changing to `CMD ["default"]` will work.

authentication/services.js
```diff
@@ -0,0 +1,100 @@
...
+});
+
+function getUserFromAccessToken(accessToken, cb) {
+  console.log("getting user for accessToken: " + accessToken);
+  authClient.users.getInfo(accessToken, function(err, user) {
+    if (err) {
+      console.log("getUser err");
+      console.log(err);
+    } else {
+      console.log("getUser success?!?");
```
tie-rack: Not sure?

authentication/services.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+    cb(user);
+  });
+}
+
+function registerAuthServices(app) {
+  app.post('/services/getUser', function(req, res) {
+    var user = getUserFromAccessToken(req.body.accessToken, function(user){
+      console.log("user is " + user);
+      res.status(200).send(user);
```
tie-rack: Is there not a failure case to consider?

data-testing/services.js
```diff
@@ -1,6 +1,8 @@
 var aws = require("./aws.js");
+var auth = require('../authentication/services.js');
 
 function registerDataVerificationServices (app) {
+  app.all('/testing/*', auth.checkJwt);
   app.post('/testing/upload', aws.uploadAddressFile);
   app.get('/testing/latest-results-file', aws.getLatestResultsFile);
```
tie-rack: Are these checked in order? It seems like `app.all('/testing/*', ...)` would match anything `app.post('/testing/upload', ...)` and `app.get('/testing/latest-results-file', ...)` are trying to match. Or are they additive? (in which case everything's fine)

Dockerfile
tank157: Ah, ok, we never try to run it without passing in something to go to the ENTRYPOINT so far, will test and change.

data-testing/services.js
tank157: They are additive. `app.all` is more or less installing `auth.checkJwt` as middleware on all matching routes, so we can avoid repeating ourselves. We were doing the same in `pg/services` until we had to change the report URLs to unauthenticated.

tank157:
@tie-rack Yeah, the duplication in `notifications/sender.js` and `data-testing/sender.js` is a known issue, and I just put in a card to refactor that when we get the chance. It'll involve some bigger structural changes (`notifications` is too feed-processing specific, so we want to change that, for example) in metis as well as changes in `batch-address-test-tool` to `s/groupName/fipsCode`. Here's the card: https://www.pivotaltracker.com/story/show/149821997

authentication/services.js
katrinamariehh: Good point--is there a reason why we're not error handling here, @tank157?  I can't quite recall.

tank157: Not really. We might try layering in the checkJwt call as middleware, then I think there'd be less need to do error handling. But if we wanted to add error handling in, then on error we could return a 401.

authentication/services.js
tank157: Wasn't at the time we wrote that log statement :)

PR #378, opened 2017-08-11
Correct the auth flow to use access token

tank157:
This is apparently the more secure way to do things, based on feedback from Auth0.

You'll need to create an API in each of the tenants,
and then you can control the expiration time of
the access token via the API settings. Take the 
audience value of the API and set it as the audience
for both AUTH0_AUDIENCE_DASHBOARD and
AUTH0_AUDIENCE_EXPRESS.

[Pivotal Card](https://www.pivotaltracker.com/story/show/150179594) to cover finishing this work.


katrinamariehh:
👍🏻

PR #379, opened 2017-08-15
Fix auth

tank157:
There are a couple of things going on now.

* We request the profile as part of the response (see change in `config.js.template`). This gives us a reduced/standardized form of the profile, so it doesn't include any of the metadata.
* The recommended way to send metadata is to use Rules to attach it to the access token. And to do so, we have to use keys that look like URIs, and should be unique. I've created rules in local/staging/prod to copy the pieces of metadata we are interested in, givenName and fipsCode, over into the access token.
* Changed `/getUser` to `/getMetadata`, since now it decodes the access token, looks up the pieces of metadata we've copied into it, and returns it to the Angular app.
* Build the local user data off of the standardized profile components and the metadata.

### DIFF
--- a/authentication/services.js
+++ b/authentication/services.js
@@ -32,29 +32,14 @@ var authClient = new AuthenticationClient({
   clientSecret: config.auth0.secret
 });
 
-function getUserFromAccessToken(accessToken, successCB, failureCB) {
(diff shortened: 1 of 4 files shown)


katrinamariehh:
👍🏻

PR #383, opened 2017-08-16
Data centralization

katrinamariehh:
PR for [this card](https://www.pivotaltracker.com/story/show/149815796).  We've added a new [page](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-0824b141d85bb91bc6b141e3a972995a) and [controller](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-e60487b9da0e5319264b2515ec453a51) on the Angular app to display a page for uploading a file as well as an [endpoint](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-a0d12f6dc47d83ed0ca74324d4e1a2d8) on the Express app to handle [uploading the file](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-cb315f8f961b41f274834e8cf3e356f8) to the data-centralization s3 bucket.

The Angular page includes a date picker which we got by adding the [angular-ui-date](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-0a08a7565aba4405282251491979bb6b) library and necessary style and script tags in [index.html](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-528c3923d718a8860f5d8c05c3931c55).  Directly using `ng-model` with the date picker to bind it's value to the `$scope` was not compatible with our version of Angular (1.2.1), so I made it work by adding a [function](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-528c3923d718a8860f5d8c05c3931c55R153) to the index page that could get the value out of the date picker element which can be called from the [controller](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-e60487b9da0e5319264b2515ec453a51R12).  Additionally, our initial plan to use a controller function to enable and disable the `submit` button also was not useable with the workaround for the date picker, so I added [another function](https://github.com/votinginfoproject/Metis/compare/data-centralization-epic...data-centralization?expand=1#diff-528c3923d718a8860f5d8c05c3931c55R162) that will spark an alert if the user attempts to submit the form without both a date and a file selected (and also does not submit the form for uploading if this happens).

There is a notion of blocking access to this area (there will be additional pages for submitting data added under this heading) from non-county users, but I'd like for us to take care of that with [this card](https://www.pivotaltracker.com/story/show/149816249) once we've determined how to handle differentiating between different types of users.

### DIFF
--- a/public/assets/css/app.css
+++ b/public/assets/css/app.css
@@ -2517,7 +2517,7 @@ form[role="search"] {
   padding: 0 1em 0 1em;
   width: 75%; }
 
-.address-file-uploaded {
+.file-uploaded {
   font-size: 1.125em;
   font-size: 1.125rem;
   font-weight: bold;
--- a/public/assets/css/app.css
+++ b/public/assets/css/app.css
@@ -1709,6 +1709,12 @@ input[type="submit"] {
   input[type="submit"]:hover {
     background: #628eb7;
     color: #fff; }
+  .button[disabled],
+  [disabled].disabled,
+  button[disabled],
+  input[disabled][type="submit"] {
+    background: #b1b3b6;
+    color: #ebebeb; }
 
 .disabled {
   background: #b1b3b6;
--- a/public/assets/js/app/app.js
+++ b/public/assets/js/app/app.js
@@ -11,7 +11,7 @@
 // ========================================================================
 
 // VIP app module with its dependencies
-var vipApp = angular.module('vipApp', ['ngTable', 'ngRoute', 'ngCookies', 'vipFilters', 'ngFileUpload', 'auth0.auth0']);
+var vipApp = angular.module('vipApp', ['ngTable', 'ngRoute', 'ngCookies', 'vipFilters', 'ngFileUpload', 'auth0.auth0', 'ui.date']);
 
 // Constants - will be added to with the properties from the external properties files
 // "vip.properties" and "map.properties"
--- a/public/assets/js/app/app.js
+++ b/public/assets/js/app/app.js
@@ -74,7 +74,7 @@ vipApp.config(['$routeProvider', '$appProperties', '$httpProvider', '$logProvide
       controller: 'AddressesCtrl'
     });
 
-    $routeProvider.when('/centralization',{
+    $routeProvider.when('/county-data/centralization',{
       templateUrl: $appProperties.contextRoot + '/app/partials/centralization/centralization.html',
       controller: 'CentralizationCtrl'
     });
--- a/public/assets/js/app/app.js
+++ b/public/assets/js/app/app.js
@@ -74,6 +74,11 @@ vipApp.config(['$routeProvider', '$appProperties', '$httpProvider', '$logProvide
       controller: 'AddressesCtrl'
     });
 
+    $routeProvider.when('/centralization',{
+      templateUrl: $appProperties.contextRoot + '/app/partials/centralization/centralization.html',
+      controller: 'CentralizationCtrl'
+    });
+
     $routeProvider.when('/feeds', {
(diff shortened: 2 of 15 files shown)


public/assets/css/app.css
```diff
@@ -1709,6 +1709,12 @@ input[type="submit"] {
   input[type="submit"]:hover {
     background: #628eb7;
     color: #fff; }
+  .button[disabled],
```
tank157: Since we're no longer disabling the button when the date/file haven't been selected, we can remove these styles.

tank157: Which, I think are not removed directly in this file, but in the scss file and then compiled into this file.

public/assets/js/app/app.js
```diff
@@ -11,7 +11,7 @@
 // ========================================================================
 
 // VIP app module with its dependencies
-var vipApp = angular.module('vipApp', ['ngTable', 'ngRoute', 'ngCookies', 'vipFilters', 'ngFileUpload', 'auth0.auth0']);
+var vipApp = angular.module('vipApp', ['ngTable', 'ngRoute', 'ngCookies', 'vipFilters', 'ngFileUpload', 'auth0.auth0', 'ui.date']);
```
tank157: This injection of `ui.date` seems to be left over from a previous attempt?

public/index.html
```diff
@@ -53,6 +53,13 @@
   <script src="assets/js/lib/auth0.js/build/auth0.js"></script>
   <script src="assets/js/lib/angular-auth0/dist/angular-auth0.js"></script>
 
+  <!-- angular-ui-date -->
+  <link rel="stylesheet" href="assets/js/lib/jquery-ui/themes/smoothness/jquery-ui.css"/>
+  <script type="text/javascript" src="assets/js/lib/jquery/jquery.js"></script>
```
tank157: Do we need to include two copies of jquery and angular.js to make this work? Or can it work with the versions we already have in previous script tags? If we have to include a second copy, that makes me think this could have all kinds of unintended side-effects from overwriting the versions we already have.

public/index.html
```diff
@@ -140,6 +150,20 @@
 
   <script src="assets/js/app/filters.js"></script>
 
+  <script>
+  function getDateValue() {
+    var dateVal = document.getElementById('datepicker').value;
+    return dateVal;
+    // var dateValEl = document.getElementById('date');
```
tank157: Cleanup commented out code.

public/assets/js/app/app.js
katrinamariehh: The ui.date is part of the library I found to use, so not something leftover.

tank157: Ah, my bad maybe on this one, I thought we only need to add it in here to configure it or use it. So we have to include modules here whether we refer to them in the code block?

tank157: Yeah, that does appear to be the case in this section.

katrinamariehh:
@tank157 I know the solution's not great, but I think what I'm more concerned about is whether or not it's going to work in IE--I can get the branch up on staging tomorrow and if you still have that VM with IE I want to see if it's performant there--that's really the thing I think we're most concerned about assuming that's what a majority of our users will use.

tank157:
Yeah, working in IE is paramount. But it's also a rather disjointed solution since the controller is relying on code far removed in the root page. I'm considering what happens to the next folks down the line if they have to modify something. Perhaps some good comments pointing out the jump out to the code in the root page in the controller would be a good idea.

PR #386, opened 2017-08-29
Display user submitted files

katrinamariehh:
For [these](https://www.pivotaltracker.com/story/show/149815923) [three](https://www.pivotaltracker.com/story/show/149817106) [cards](https://www.pivotaltracker.com/story/show/149816024) adding a table on the data centralization page to show files that have been submitted by a user.  In the case of a county user, they will see their own files; in the case of a state admin user, they will see the files submitted by all counties in their state with an additional field showing the county that submitted each file; and in the case of a super admin user, they will see all files submitted by all users with both state and county identified for each file.  The different fields to be displayed in each instance are achieved using the `hasRole()` function we added in the initial data-centralization work.

The form to upload data is only displayed to county users at this time(this can be toggled if we want to display it for them, but I don't think that's the intention).  We're also identifying the states by name based on their FIPS code, this could be expanded to identify counties as well instead of just showing FIPS codes.

I also amended the `if` statement around showing the county data upload button based on having one of the three aforementioned roles (`data-centralization`, `state-admin`, or `super-admin`).  We also updated the version of the 'aws-sdk' and updated the `listObject` functions accordingly.


public/app/partials/centralization/centralization.html
```diff
@@ -6,16 +6,40 @@
...
+      <br>
+      <div class="button" ng-click="submit()">Submit</div>
+    </form>
+  </div>
 
+  <br>
+  <div ng-if="submittedFiles.length > 0">
+  <label ng-if="hasRole('data-centralization')">Your previously submitted files:</label>
+  <label ng-if="hasRole('state-admin')">Files submitted by counties in your state:</label>
+  <label ng-if="hasRole('super-admin')">Files submitted by all users:</label>
```
tank157: This could get messy if someone is given more than one of these roles. Is there any kind of `else-if` capabilities within angular with which we can prevent multiple labels from showing up at once?

public/app/partials/centralization/centralization.html
```diff
@@ -6,16 +6,40 @@
...
+      <tr>
+        <th ng-if="hasRole('super-admin')">State</th>
+        <th ng-if="hasRole('super-admin') || hasRole('state-admin')">County</th>
+        <th>File name</th>
+        <th>Election date</th>
+        <th>Last modified</th>
+      </tr>
+      <tr ng-repeat="file in submittedFiles">
+        <td ng-if="hasRole('super-admin')">{{getStateFromFips(file.stateFips)}}</td>
+        <td ng-if="hasRole('super-admin') || hasRole('state-admin')">{{file.countyFips}}</td>
```
tank157: We talked about getting county names from a JSON or other data structure. Just wondering if that's something Maria was ok leaving out and just using county FIPS instead of county name here?

public/app/partials/centralization/centralization.html
katrinamariehh: That's certainly a reasonable suggestion, but is there a reason why we would give someone multiple roles?

public/app/partials/centralization/centralization.html
katrinamariehh: This was not an explicit ask in the card; I'll confirm with her in #vip if this is something she wants us to change before deploying.

katrinamariehh: Maria is fine with leaving things they way they currently are.

public/app/partials/centralization/centralization.html
tank157: It's not so much a reason why we would intentionally do that as it is a very real possibility someone will do it at some point after us, since the way of setting up roles is very fluid and process based rather than enforced by an interface. Just experience tells me that if someone can set it up wrong, they likely will. It's not a huge concern, since this doesn't seriously affect the functionality, and I was more interested if angular did provide some mechanism for handling this.

public/app/partials/centralization/centralization.html
tank157: 👍

public/app/partials/centralization/centralization.html
katrinamariehh: I think I'm just hesitant here because it means we have to accommodate all 7 possible combinations of these 3 roles to determine what title we should give the table, which feels like overkill.

tank157: I was thinking we'd just see if someone had `super-admin` first and give them the most open interpretation (so even if they had `state-admin`, we wouldn't give them the label for state admins), then once we've handled any users with `super-admin`, we'd check for `state-admin`, and then `data-centralization` would more or less be the default case.

Something like:
``` if(hasRole('super-admin')) {
  // display super-admin label
 } else if {hasRole('state-admin')) {
  // display state-admin label
 } else {
  // display data-centralization label
 }```

Comment (katrinamariehh) [reply public/app/partials/centralization/centralization.html]: This all feels like overkill; I'll just make the title of the table more generic.

File: public/app/partials/centralization/centralization.html [hunk 3]
```diff
@@ -22,9 +22,7 @@
 
   <br>
   <div ng-if="submittedFiles.length > 0">
-  <label ng-if="hasRole('data-centralization')">Your previously submitted files:</label>
-  <label ng-if="hasRole('state-admin')">Files submitted by counties in your state:</label>
-  <label ng-if="hasRole('super-admin')">Files submitted by all users:</label>
+  <label>Previously submitted files:</label>
```
tank157: 🚢

PR #387, opened 2017-08-30
Data centralization epic

katrinamariehh:
PR to deploy all the work in [this epic](https://www.pivotaltracker.com/epic/show/3681391); the individual branches have all been PR'd and merged into here so this just needs a cursory overview to make sure nothing stands out before deployment.

It should be noted as well that most of this will be invisible until we start making users with the roles we defined as part of this work; i.e. what we have defined [here](https://github.com/votinginfoproject/wiki/blob/user-roles/auth0/user-roles.md).  The link at the top of the page to the county data uploads section is invisible unless one of the three roles is assigned (`data-centralization`, `state-admin`, or `super-admin`), so until users with those roles are created or existing users are given those roles, this will all be invisible when deployed to production.


public/index.html
```diff
@@ -140,6 +148,17 @@
 
   <script src="assets/js/app/filters.js"></script>
 
+  <script>
+  function getDateValue() {
```
katrinamariehh: This should have some explanation; there are comments in the controller code where it's used, but I'll add some explanation here as well.

PR #388, opened 2017-08-31
Ballot measure contest errors

katrinamariehh:
For [this bug card](https://www.pivotaltracker.com/story/show/149515783) this PR adds functions to `v5_1_queries.js` and `csv.js` to allow errors to be requested using both an element type (as pulled from an ltree path) as well as with a scope; this will solve the specific problem of ballot measure contest errors not getting pulled into their scoped error page and report.  I made it generic in case there are other error types that follow this same pattern.

The issue at hand is that the ballot measure contest errors in question (UTF-8 errors in the text) are saved as global errors rather than as errors established after the values have been added to `xml_tree_values`; so when they're added to `xml_tree_validations` they don't have `path` which is required for the original way the summary page and report were generated.  These changes allow the summary error page and report to be generated with both an `elementType` which works for errors that have an ltree path as well as a `scope`, which is all we have in the case of these global errors to determine where the error should be shown.

It should be noted that this does not impact the main errors page (and I've got approval from Maria to leave that part alone for now) which will show the incorrect number of errors.  There is a single function in `data-processor` that generates the data to populate the main error and updating that to pull these global errors for their related elements is a more significant undertaking that we'll put aside for now.  There is a Michigan feed noted in the bug card that can used to test this locally.

### DIFF
--- a/pg/csv.js
+++ b/pg/csv.js
@@ -85,7 +85,7 @@ var scopedXmlTreeValidationQuery = function(elementTypes) {
  WHERE r.public_id = $1 AND v.path ~ 'VipObject.0." + elementTypes.join("|") + ".*'";
 }
 
-var ballotMeasureContestScopedXmlTreeValidationQuery = function() {
+var elementTypeAndScopeXmlTreeValidationQuery = function(elementTypes, scope) {
   return "SELECT v.severity, v.scope, v.path, v.parent_element_id AS identifier, \
         v.error_type, v.error_data \
  FROM xml_tree_validations v \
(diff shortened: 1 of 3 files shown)


pg/csv.js
```diff
@@ -85,6 +85,48 @@ var scopedXmlTreeValidationQuery = function(elementTypes) {
...
+var elementTypeAndScopeXmlTreeValidationQuery = function(elementType, scope) {
+  return "SELECT v.severity, v.scope, v.path, v.parent_element_id AS identifier, \
+        v.error_type, v.error_data \
+ FROM xml_tree_validations v \
+ INNER JOIN results r ON r.id = v.results_id \
+ WHERE r.public_id = $1 AND (v.path ~ 'VipObject.0."+ elementType + ".*' or v.scope like '" + scope + "%')";
+}
+
+var elementTypeAndScopeXmlTreeValidationErrorReport = function(elementType, scope) {
+  // var elementTypes = Array.prototype.slice.call(arguments);
```
tank157: Let's clean up this commented out code and the same on line 109.

katrinamariehh:
I have an idea about how to get the number to show up correctly on the main feed summary page (since it's relatively similar to what we had to do to break out the polling locations), but it's taking a bit longer than expected so I think what I'll do is get this merged in and worry about fixing the main summary page later.  Does that work for you @tank157?

PR #391, opened 2017-09-12
Notifications refactor

katrinamariehh:
This PR is for [this card](https://www.pivotaltracker.com/story/show/149821997) about re-factoring our notifications-related code to be better organized and use appropriate language (i.e. getting away from "group" which was a Stormpath-dictated term and instead referring to FIPS codes where appropriate.

This refactor moves all the aws-related code into a single file under the aws directory with a services file for accessing it (`aws/s3.js` and `aws/services.js`).  This includes functionality that sends and retrieves files for the batch address test tool and for the data centralization tool.

All notifications-related code is now under the `notifications` directory.  Underneath that `sender.js` and `services.js` contain all of the more generalized code for sending notifications with specific `content` and `queue` files under the `data-testing` and `feed-processing` sub-directories.  Any future additional notifications should fit well here; new sub-directories are easily created and the `messageOptions` in `sender.js` can be updated to fit as needed.

The last piece yet to be done for this to be deployed is to also update the `batch-address-test-tool` to send messages with a fips code rather than a group name.  I'll link when that PR is ready.


notifications/sender.js
```diff
@@ -96,19 +116,36 @@ module.exports = {
...
     }
+  },
+  sendDataTestingNotifications: function(message) {
+    var messageType  = (message['status'] == "ok") ? messageOptions['testingComplete'] : messageOptions['errorDuringTesting'] ;
+    var fipsCode = message["fipsCode"];
+    if (fipsCode === undefined) {
+      logger.warning("No fips code in batch-address.file.complete message.  Can't send batch address testing finished email notification.");
+      logger.info(message);
+    } else if (fipsCode === "undefined") {
+      if (config.email.adminGroup === undefined || config.email.adminGroup === null) {
```
tank157: How do you feel about the continued use of `adminGroup` here and in the other notification sending function? It sort of carries the "group" notion forward, but in this case there's no admin fipscode, so maybe it's ok? Did you consider renaming it to anything else?

katrinamariehh: I think I thought about this briefly, but yeah, since there's not admin fipscode group actually kind of works here though it does continue the use of the word "group".  Maybe it should be `adminList` or something like that?

tank157: At this point group is as good as list. Just brainstorming other ideas and none really popped out.

PR #396, opened 2017-10-03
Only need to set hasRole in rootScope once if we do it correctly

tank157:
Refactor where we set hasRole to the rootScope. This way we only have to do it one place, and don't miss other pages we might add later.


katrinamariehh:
👍🏻😎

PR #397, opened 2017-10-03
Save elections

katrinamariehh:
For [this card](https://www.pivotaltracker.com/story/show/150868130).

The save elections form now requires both fields to be filled in before it enables the button to save it, it requires the user to select a date that is in the future, and creates an election in the database by submitting a request to the node backend.  We've also added migrations for the new database that will save dashboard-specific data.


src/cljs/early_vote_site/election/events.cljs
```diff
@@ -0,0 +1,47 @@
...
+  {:state_fips (get-in db [:elections :form :state])
+   :election_date (get-in db [:elections :form :date])})
+
+(re-frame/reg-event-fx
+ :election-form/save
+ (fn [{:keys [db]} _]
+   (let [data (create-params db)]
+     {:db (assoc db :xhr-spinner true)
+      :http-xhrio {:method          :post
+                   :uri             "http://localhost:4000/earlyvote/elections"
```
tie-rack: This will have to not be hardcoded... Could something like this work?

```clj
:uri    (-> js/window .-location .-origin (str "/earlyvote/elections"))
```

src/cljs/early_vote_site/election/subs.cljs
```diff
@@ -0,0 +1,20 @@
...
+ (fn [db]
+   (some-> db
+           (get-in [:elections :form :date])
+           .getTime)))
+
+(re-frame/reg-sub
+ :create-disabled?
+ (fn [db]
+   (or (= nil (get-in db [:elections :form :date]))
+       (= "" (get-in db [:elections :form :state])))))
```
tie-rack: Minor thing: this feels just too specific (that `date` would be nil but `state` would be the empty string). `clojure.string/blank?` covers both and using it wouldn't feel like relying on special knowledge of how those are initialized.

public/early_vote.html
```diff
@@ -2,13 +2,36 @@
...
+    <div class="main" role="main">
+      <div class="wrapper">
+        <div class="content">
+
+          <div id="app"></div>
+          <script src="assets/js/compiled/early_vote.js"></script>
+          <script>early_vote_site.core.init();</script>
+        </div>
+        <footer role="contentinfo">
+          <small>&copy; Copyright 2014 Voting Information Project.</small>
```
tie-rack: Super minor: it's 2017! (I don't want to believe it either)

Gruntfile.js
```diff
@@ -133,7 +143,7 @@ module.exports = function(grunt) {
 
   grunt.loadNpmTasks('grunt-bower-task');
 
-  grunt.registerTask('default', ['concurrent:dev_configure','concurrent:dev_run']);
+  grunt.registerTask('default', ['concurrent:dev_configure', 'concurrent:migrate', 'concurrent:dev_run']);
```
tie-rack: Does `concurrent:migrate` need to be added to the staging and production pipelines?

early-vote/elections.js
```diff
@@ -0,0 +1,21 @@
...
+  var params = [];
+  params.push(uuidv4());
+  params.push(decodeURIComponent(req.body['state_fips']));
+  params.push(decodeURIComponent(req.body['election_date']));
+  return params;
+}
+var createElection = util.simpleCommandResponser(createSql, createElectionsParamFn);
+
+function registerElectionServices (app) {
+  //app.all('/evs/elections/*', auth.checkJwt);
```
tie-rack: Is this here as a reminder for later (when we need a JWT involved)? Or can this go?

src/cljs/early_vote_site/election/views.cljs
```diff
@@ -0,0 +1,69 @@
...
+                               {:min-date (js/Date.)
+                                :on-select #(re-frame/dispatch
+                                              [:election-form/date-selected %])}}]]
+      [:button.button {:on-click #(re-frame/dispatch [:election-form/save])
+                       :disabled @(re-frame/subscribe [:create-disabled?])}
+       "add"]]
+      [:p @state]
+      [:p (some-> @date (js/Date.) .toString)]]))
+
+(def election-list
```
tie-rack: Just want to make sure I've got this right: The actual code for showing real elections as they've been entered is in a later story?

katrinamariehh: Yep!  That's what's coming up next.

src/cljs/early_vote_site/election/views.cljs
```diff
@@ -0,0 +1,69 @@
...
+      [:td {:name "election-state"} "New York"]
+      [:td {:name "election-date"} "Nov 7, 2017"]]
+     [:tr
+      [:td {:name "election-state"} "Colorado"]
+      [:td {:name "election-date"} "Nov 7, 2017"]]]])
+
+(defn election-list-row
+  ; this will need to take in data from the database to construct the list of elections
+  ; what's the best entry point for this?  do we put something in db.cljs that
+  ; can pull things from the Postgres database?
```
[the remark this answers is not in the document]
katrinamariehh: Yeah, this was more a note for later when we are pulling live data that I know it needs to go here but I'm not quite sure how to do it.  I think you're right though, that with the other work Troy put in that we'll be putting an AJAX request here.

src/cljs/early_vote_site/election/events.cljs
tank157: I was going to push this off a little longer, as this was the only way that worked for a figwheel setup short of pulling in configuration. But we haven't really been relying on figwheel much anymore, so I'll see what I can do.

(In the figwheel setup, node is still running at 4000 and serves the backend api, but the browser is on 3449 for figwheel, hence the reason just using location origin wouldn't have worked).

src/cljs/early_vote_site/election/subs.cljs
tank157: Good call. React doesn't like you supplying null/nil for a select value, so it became the empty string, but `blank?` would be cleaner.

Gruntfile.js
tank157: Yeah, I was waiting until we got to a staging/prod deploy first, and since we're actually gearing up for a staging deploy soon, now would be a good time to add that.

early-vote/elections.js
tank157: It's here as a reminder, we've not yet added authentication integration to the early vote site app, so it can't send them back to the backend yet.

tank157:
@tie-rack I think we've now addressed everything if you want to take a look again when you get a chance. Thanks!

PR #398, opened 2017-10-05
Ie js fix

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/151721567)

This fixes some javascript compatibility errors and makes some changes to authentication needed for IE 10. I checked in IE 10 and Safari, would be good to check other browsers. IE 9 I don't think will work with the new authentication anymore without maybe a lot more work, but it's now out of our 3 browser versions support.


katrinamariehh:
Just to make sure, is this up on staging so I can test it out on a Windows machine?  I want to make sure I do my due diligence in reviewing it!

tank157:
Yup! I think everything but perhaps the last commit is up on staging now. The last commit just removed a `console.log(...)` statement so it should be fine to test without that deployed (famous last words?)

katrinamariehh:
I checked this on Windows 10 with IE and Edge, waiting to confirm it's OK on the setup specified in the bug card (Windows 7, IE 10).

katrinamariehh:
Per Maria this is good to go on Windows 7/IE 10. 🐋

PR #401, opened 2017-11-09
Early Vote Site Tool Epic PR

tank157:


tank157:
Basically this now covers most/all the work for [this epic](https://www.pivotaltracker.com/epic/show/3714455). The cards that are open have been annotated to point out where in the code we implement these features.

dashboard/util.js
```diff
@@ -0,0 +1,133 @@
...
+          var params = paramFns[i](req);
+          if (params) {
+            for (var j = 0; j < params.length; j++) {
+              ret.push(params[j]);
+            }
+          }
+        }
+      return ret;
+    }
+  },
```
tie-rack: These extractors are so nearly `map` and `mapcat` they _could_ be:

```js
queryParamExtractor: params => req => params.map(param => decodeURIComponent(req.query[param]))
pathParamExtractor: params => req => params.map(param => decodeURIComponent(req.params[param]))
bodyParamExtractor: params => req => params.map(param => req.body[param])
compoundParamExtractor: paramFns => req => paramFns.reduce((allParams, paramFn) => allParams.concat(paramFn(req)), [])
```

Maybe you don't have to go that far, but using `map` in place of all the `for (var i = 0; i < params.length; i++)` bits is worth not reinventing `Array.prototype` functions.

early-vote/assignments.js
```diff
@@ -0,0 +1,39 @@
...
+                                           util.pathParamExtractor(['assignmentid']));
+
+//create assignment
+var createSql = "INSERT into assignments(id, early_vote_site_id, schedule_id) " +
+  "values ($1, $2, $3);"
+
+var createParamsFn =
+  util.compoundParamExtractor([util.uuidGenerator(),
+                               util.pathParamExtractor(['earlyvotesiteid']),
+                               util.bodyParamExtractor(['schedule_id'])]);
```
tie-rack: So _functional_.

early-vote/file-generation.js
```diff
@@ -0,0 +1,309 @@
...
+/**
+ * Given an array of schedules, creates a map where the schedules are
+ * keyed by id for easy lookup.
+ */
+var schedulesById = function(schedules) {
+  var byId = {};
+  schedules.forEach(function(schedule) {
+    byId[schedule["id"]] = schedule;
+  });
+  return byId;
```
tie-rack: This function and the next two have that classic `reduce` look to them.

tank157:
@tie-rack Ok, I've taken your suggestions and run with them. Had some difficulty in some of the reduces, had to revert to `hash[key] = newval` syntax that I wasn't as thrilled with but trying to use the new `Map` type didn't work out, nor did other things I tried. Anyway, the arrow functions are neat, I'd only been vaguely aware of them, so getting pointed to actual documentation on them helped a lot!

public/assets/css/re-frame.css
```diff
@@ -1,4 +1,4 @@
-/* variables-and-mixins.scss */
+breadcrumb/* variables-and-mixins.scss */
```
tie-rack: This `breadcrumb` looks out of place

PR #402, opened 2017-11-15
update build/deploy scripts to work with newer agents

tank157:
Update build/deploy scripts to work with new agents.

Primarily we drop the PEM_FILE requirement, and a few other tweaks so builds and artifacts savings will be run on all branches (so staging _can_ be deployed via BK as well).

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,13 +119,19 @@ the Angular config file, you can return to using `node app.js` if you like.
 
 ## Deploying the Project
 
+### Automatic
+
+Merges to `master` branch will be deployed to production automatically.
+
+### Manual
+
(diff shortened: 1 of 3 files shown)


tank157:
@tie-rack I've got the BuildKite pipeline for this partially updated to make the transition. The old steps are now renamed with "legacy" and the new steps are in place. Deployment is current restricted in both legacy and new by branch filtering on NONE, but the plan will be to only selectively deploy to staging, and turn master branch filtering on deploy to production.

Once this is approved, I'll remove legacy steps, turn master branch deploys on production, and merge this.

PR #409, opened 2018-03-15
V3 file generation

tank157:
Adds support to generate the 3.0 spec `early_vote_site.txt` file, and generate the custom days_times_open with 0 to many schedules.

[Pivotal Card](https://www.pivotaltracker.com/story/show/155953862)


tank157:
Re-opening it to be based off the auth0-upgrade branch.

PR #412, opened 2018-04-11
Admin notifications

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/152721498)

Adds Slack notifications for feed processing conclusion events and batch address conclusion events as well. By which I mean when a feed or batch address file finishes processing, and sends a rabbitmq message back to Metis, this figures out the status of the event and posts a message to a configured Slack channel with that info. So it's handling successes as well as many failure conditions.

### DIFF
--- a/config.js
+++ b/config.js
@@ -79,4 +79,12 @@ config.earlyVoteSites = {
   bucket: process.env.EARLY_VOTE_SITES_BUCKET_NAME
 }
 
+config.slack = {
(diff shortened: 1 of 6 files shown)


tank157:
Yeah, we used to send some of this via email, but a lot of that went away when we transferred to Auth0. And it's not like we couldn't re-implement the emails with some effort, but frankly this way we skip another point of Auth0 integration/dependency (ie searching for all users that are `super-admin`) and Slack >> email.

PR #414, opened 2018-06-04
Dependency Updates per Casaba's Security Review

ddellacosta:
https://www.pivotaltracker.com/story/show/157015598

Note that this will require updating [clojure-and-node-docker](https://github.com/democracyworks/clojure-and-node-docker) to provide the version of Node used here (8.11.2) before it can be deployed. Update: [related PR](https://github.com/democracyworks/clojure-and-node-docker/pull/1).

This has only been tested locally, but appears to work as before (yay no tests).

Addendum: since we are still testing this on staging, I just dropped in the "remove tests" changes from [this card](https://www.pivotaltracker.com/story/show/158093561) here, which may as well have just been in this PR anyways.


app.js
```diff
@@ -42,27 +51,28 @@ var redirectHttps = function(req, res, next) {
 };
 
 // all environments
-app.use(express.compress());
-app.use(express.favicon(config.web.favicon));
-app.use(express.logger(config.web.loglevel));
+app.use(compress());
+app.use(favicon(config.web.favicon));
+//app.use(morgan(config.web.loglevel));
```
tank157: Do we need this commented out morgan line?

app.js
```diff
@@ -42,27 +51,28 @@ var redirectHttps = function(req, res, next) {
...
+app.use(methodOverride());
+app.use(cookieParser());
 
 // Redirect non-https load balanced clients to https
 app.use(redirectHttps);
 // /ping for load balancer health checking
 app.get('/ping', function (req, res, next) { res.send('pong'); });
 
-app.use(app.router);
+//app.use(app.router);
```
tank157: Seems like another commented out line we can remove

public/assets/css/app.css
```diff
@@ -1496,15 +1496,15 @@ aside {
   @media (min-width: 1116px) {
     aside {
       float: right;
-      margin-left: 3.7037%;
-      width: 30.8642%;
+      margin-left: 3.7037037037%;
+      width: 30.8641975309%;
```
tank157: Just curious where the extra degrees of precision came from for this file? They seem pretty specific.

tie-rack: Future-proofing for 225,000 DPI screens?

ddellacosta: This is auto-generated as far as I know--I was surprised that we had build files committed actually--so I can't really tell you, but I assume it's whatever sass is doing during building.

app.js
ddellacosta: Good catch on this and the other one! I also have to see if we want to configure the logs to use the shorter format, that's something that got changed as a result of updating express.

public/assets/css/app.css
tank157: Ok, I thought it might be as such, was still just a little strange to see :)

ddellacosta: Agreed, gave me a chuckle when I saw it actually.

PR #415, opened 2018-06-20
add county fips to 5.1 and 3.0 output

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/158472505)


tank157:
[polling_place-4.txt](https://github.com/votinginfoproject/Metis/files/2121253/polling_place-4.txt)
[early_vote_site-14.txt](https://github.com/votinginfoproject/Metis/files/2121254/early_vote_site-14.txt)

Two sample files from the same "election" showing that the county_fips is added to the end of the output for both 3.0 and 5.1 style files.

PR #417, opened 2018-06-21
Adds some extra handling for authentication

ddellacosta:
https://www.pivotaltracker.com/story/show/157016344

I've done a few things that possibly overlap here.

- it appears that there are cases where we are clearly not authenticated but the checks we're making don't catch this and do not always redirect--I've added a fall-through case here to ensure we redirect, but...

- this is obviously still not sufficient because the `CentralizationCtrl` controller doesn't seem to be affected. So I've added some special handling there.

I don't really love what I've done here, but the alternative is to spend a bunch of time on this figuring out The Right Way© according to Angular 1.6 and rewrite chunks of Metis and I think it's really not worth it. Hopefully this is enough to catch the remaining holes (it seems to be working fine for all the feed pages I've checked) and otherwise we can play whack-a-mole with any others that pop up.


ddellacosta:
@tank157 Oh, also, note that I've used the as-yet-unmerged dependency upgrade branch from [here](https://github.com/votinginfoproject/Metis/pull/414) as it made no sense to base this off of the older Angular in master.

tank157:
@ddellacosta i think you can change the PR to merge into that dependency upgrade branch instead, and then it'll just show the relevant commits. I've done that a few times when I've branched new work off of as-yet-unmerged other work.

tank157:
But otherwise, those last two commits look fine to me. And I was expecting much more hackery based on your comments, but these seemed relatively clean. LGTM 🐍

PR #420, opened 2018-08-02
Secure early vote site tool

tank157:
This work should replace the spots where we sent information like fips codes and roles from the browser to the backend, and also implements a layer of data integrity checks such that users authorized for one state cannot send the resource id of another state and have it succeed. It also locks down the things that different roles are permitted to perform to just the relevant data.

[Pivotal Card](https://www.pivotaltracker.com/story/show/157016309)


early-vote/access.js
```diff
@@ -0,0 +1,262 @@
...
+ 			}
+ 		}
+ 	};
+
+ 	conn.query(function(client) {
+ 		client.query(sqlCommand, params, callback);
+ 	});
+ }
+
+var verifyElectionSql = `select * from elections where id = $1 AND state_fips in ($2);`
```
ddellacosta: Is there any (performance) value in refining the list of columns returned on these queries?

early-vote/access.js
```diff
@@ -0,0 +1,262 @@
...
+var verifyElectionParamFn = util.compoundParamExtractor([util.pathParamExtractor(['electionid']),
+																								 		 		 stateFipsExtractor()]);
+/**
+ * Express middleware that checks if user is super admin or that the election
+ * matches their state fips.
+ */
+function verifyElection(req, res, next) {
```
ddellacosta: Not sure if it's worth it, but seems like these validation functions all follow the same pattern other than some naming particulars, maybe worth refactoring into a single fn with some parameters?

Or not.

early-vote/access.js
tank157: Not a bad idea, I could just pull IDs and get the results I want and maybe a bit better performance

early-vote/access.js
tank157: I think I tried going down that road initially but it got kind of confusing quick.

early-vote/access.js
ddellacosta: Oh also--is whatever library we're using doing prepared statements properly so we don't have to worry about SQL injection attacks? ...or are we otherwise validating these arguments?

PR #422, opened 2018-09-13
give super-admins the ability to filter files

tank157:
There's no Pivotal Card for this, it's an outcome of looking into some issues we're seeing with the Data Centralization file list for super-admins.

Allows super-admin users to pick a state to
filter the files by that state, taking advantage
of the prefixing we've done.

Since we've hit a 1000 file limit, this was the
quickest way for now to both introduce a useful
feature and to give us some breathing room while
we consider a more stable long term solution.

### DIFF
--- a/aws/s3.js
+++ b/aws/s3.js
@@ -143,7 +143,11 @@ module.exports = {
     var bucketName = centralizationBucket;
 
     if (auth.isSuperAdmin(req)){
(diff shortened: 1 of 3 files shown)


tank157:
Yeah, long term would be something like paging, and/or removal of old data (some history is maybe good, but maybe not keeping everything forever)

PR #437, opened 2019-04-11
Use the county fips/name data to build a dropdown

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/165226317)

### DIFF
--- a/src/cljs/early_vote_site/places.cljs
+++ b/src/cljs/early_vote_site/places.cljs
@@ -3202,30 +3202,27 @@
   ([fips]
    (get fips->name fips fips)))
 
-(defn county-fips?
-  "Returns true if the fips looks like a county fips (has 5 characters)"
-  [fips]
-  (= 5 (count fips)))
-
-(defn in-state?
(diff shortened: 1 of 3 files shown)


src/cljs/early_vote_site/places.cljs
```diff
@@ -3201,6 +3202,31 @@
   ([fips]
    (get fips->name fips fips)))
 
+(defn county-fips?
+  "Returns true if the fips looks like a county fips (has 5 characters)"
```
sduckett: It looks like a FIPS is usually digits; does it matter how specific we get here?

tank157: I thought about saying "5 digits" but the code doesn't specifically check for it, nor does it really need to. The source data set (a map) consists of keys for both state and county fips, so it's really about being able to tell if a key looks like a county fips (5 characters) over a state fips (2).

tank157: Come to think of it, these would do well inside a `letfn` in this case. That'll signify they aren't generic functions, but specific to this case.

PR #440, opened 2019-05-09
fixes file upload bug

lizstarin:
Uploading a file was causing Metis to crash in some browsers (not totally clear on which ones). Maybe a problem with `multiparty.form.parse`?


lizstarin:
https://www.pivotaltracker.com/story/show/165918150

lizstarin:
Yeah, I was thinking the same about Clojure when I got to that part.

PR #441, opened 2019-05-09
display upload progress to user

lizstarin:
Shows file upload progress with both a progress bar and text. (These both disappear when the upload is complete.)

upload progress bar


lizstarin:
Note: for small files/quick speeds, the bar appears very briefly, at 100%. I don't think it looks bad (at least in my browser), but maybe a little delay before showing the progress bar would be good?

lizstarin:
https://www.pivotaltracker.com/story/show/165312787

PR #444, opened 2019-05-30
Adds date to VIP notifications

lizstarin:


lizstarin:
https://www.pivotaltracker.com/story/show/166125927

tank157:
Hmmmm, i'm not understanding why there are all these extra commits showing up here. I thought it should be actually be merging into `staging` instead of `master`, but when I made that change to the PR settings, there were still a fair number of extra commits showing up, so I reverted it back for the moment until we can look into this. But we probably don't want to merge this into `master` directly, as it'll end up deploying some part of the new upload functionality when we do, and I think there's more stuff accumulated in the `staging` branch that also needs to go. Might need to try a rebase on this.

lizstarin:
Aahhh, I probably branched from the wrong place.

PR #447, opened 2019-06-24
Wabbit season

tank157:
Removes the use of rabbitmq from the application and replaces it with various AWS services (SQS mostly). A lot of refactoring and then tweaks to get it operating correctly in the staging environment. Tackled a number of configuration streamlines and version upgrades as well. Oh, and removed the not really used/useful newrelic integration as well.

[Pivotal card](https://www.pivotaltracker.com/story/show/166622758)

Pairs with branches of the same name in `data-processor` and `batch-address-test-tool`.


aws/s3.js
```diff
@@ -25,14 +27,13 @@ module.exports = {
       res.writeHead(200, {'content-type': 'text/plain'});
       res.write('received upload:\n\n');
       res.end(JSON.stringify(files));
-      var fileStream = fs.createReadStream(files.file.path);
+      var fileStream = fs.createReadStream(files.file[0].path);
```
tank157: Opportunistically fixing a current bug here, something changed with the format of the form parsing and now we need to use the array indexing to access the file being uploaded.

aws/s3.js
```diff
@@ -25,14 +27,13 @@ module.exports = {
...
       fileStream.on('error', function (err) {
         if (err) { throw err; }
       });
       fileStream.on('open', function () {
-        var s3 = new AWS.S3();
         var bucketName = batchAddressBucket;
-        var fileName = fipsCode + '/input/' + files.file.originalFilename;
+        var fileName = fipsCode + '/input/' + files.file[0].originalFilename;
```
tank157: Opportunistically fixing a current bug here, something changed with the format of the form parsing and now we need to use the array indexing to access the file being uploaded.

aws/s3.js
```diff
@@ -117,7 +116,7 @@ module.exports = {
       for (var i = 0; i < files.length; i++) {
         var key = files[i]["Key"];
         var lastModified = files[i]["LastModified"];
-        var stateFips = key.match(/(\d{2})\//);
+        var stateFips = key.match(/^(\d{2})\//);
```
tank157: Another opportunistic fix here, to anchor the optional FIPS detection to the start of the string only, otherwise the regex can catch the end of the date as a FIPS code.

public/assets/js/app/controllers/asideController.js
```diff
@@ -20,7 +20,7 @@ function AsideCtrl($scope, $rootScope, $http, $feedDataPaths, $routeParams) {
 
   function sendEmail(election) {
     $http.post('/notifications/approve-feed', {
-      ":public-id": feedid,
+      publicId: feedid,
```
sduckett: 👏

metis@.service.template
```diff
@@ -31,22 +29,19 @@ ExecStartPre=/bin/bash -c 'echo DB_PORT_5432_TCP_PORT="$(curl -s http://${COREOS
 ExecStartPre=/bin/bash -c 'echo DB_ENV_POSTGRES_USER="$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/data-suite/postgres/username?raw)" >> /tmp/${CONTAINER}--env'
 ExecStartPre=/bin/bash -c 'echo DB_ENV_POSTGRES_PASSWORD="$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/data-suite/postgres/password?raw)" >> /tmp/${CONTAINER}--env'
 ExecStartPre=/bin/bash -c 'echo DB_ENV_POSTGRES_DATABASE="$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/data-suite/postgres/database?raw)" >> /tmp/${CONTAINER}--env'
-ExecStartPre=/bin/bash -c 'echo VIP_DP_RABBITMQ_EXCHANGE="$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/data-suite/rabbitmq/exchange?raw)" >> /tmp/${CONTAINER}--env'
 ExecStartPre=/bin/bash -c 'echo VIP_DP_SES_FROM="$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/data-suite/ses/from-address?raw)" >> /tmp/${CONTAINER}--env'
-ExecStartPre=/bin/bash -c 'echo VIP_DP_SES_REGION="$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/data-suite/ses/region?raw)" >> /tmp/${CONTAINER}--env'
```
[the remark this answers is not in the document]
tank157: I just consolidated every region into a singular setting as I don't see a case for using discrete regions for different services, especially not with the same credentials.

app.js
```diff
@@ -21,7 +21,7 @@ var errorhandler = require('errorhandler');
 var http = require('http');
 var path = require('path');
 var fs = require('fs');
-var queue = require('./queue');
```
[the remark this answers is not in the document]
tank157: That's a good point, let me reflect on this. It wasn't so much an intentional move away from this name as it was a progressive development where eventually I'd gutted all the code out of `queue`.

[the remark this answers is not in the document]
tank157: I think I'll hold off on changing this for now. Mostly because my longer term plans include removing all this from Metis into a separate notifications service written in clojure as part of our refactoring away of Metis anyway.

PR #449, opened 2019-07-25
Dasher elections

lizstarin:
Goes through `dasher` for all CRUD functions for `election`s.


dasher/services.js
```diff
@@ -71,6 +71,84 @@ function registerDasherServices(app) {
...
+        res.status(500).send(error);
+      } else if(response && response.statusCode){
+        res.status(response.statusCode).send(response.body);
+      }
+    })
+  });
+
+  // update election through dasher
+  app.put('/dasher/elections/:id', auth.checkJwt, function(req, res) {
+    var options = {
```
tank157: Seeing this same function structure over and over with only small differences makes me think about writing a function to encapsulate it, but then I recall this is only a stop gap measure and so it's probably fine to leave as is since we'll remove it before too long (I hope!)

public/assets/js/app/controllers/profileController.js
```diff
@@ -21,4 +21,12 @@ function ProfileCtrl($scope, $rootScope, $route, $location, $dasherService, $aut
   $scope.generateApiKey = function(){
     $dasherService.generateApiKey($scope.apiKey);
   };
+
```
tank157: These invocation spots are just for local testing, correct? We probably don't want local test work in the PR.

public/index.html
```diff
@@ -245,8 +245,14 @@ <h1 id="pageHeader-title">{{pageHeader.title}}</h1>
         </div>
       </div>
 
+      <div ng-controller="ProfileCtrl">
```
tank157: Same here if these are just for testing something out locally we don't really want it in the PR.

public/assets/js/app/controllers/profileController.js
lizstarin: Oh my, yes. No way that's going in production.

lizstarin:
Would it be easier to make a separate dev-only view to hold the buttons? However, I think it's fine to just delete them. I built them when I hadn't yet integrated the new routes into the app (and wasn't sure when that would happen).

PR #451, opened 2019-08-14
Fix ev queries

tank157:
State Admins can't unassign/delete schedules because the query to verify that they should have access was incorrect. So this fixes that query and uses a consistent alias for the assignments table in the `access.js` file. Also, when a schedule was deleted, it said the Early Vote Site was deleted in the flash message, so that is also fixed here.

### DIFF
--- a/src/cljs/early_vote_site/early_vote_site_detail/events.cljs
+++ b/src/cljs/early_vote_site/early_vote_site_detail/events.cljs
@@ -198,7 +198,7 @@
 (defn delete-schedule-success
   [{:keys [db]} [_ early-vote-site-id]]
   {:db db
-   :dispatch-n [[:flash/message "Early vote site deleted"]
+   :dispatch-n [[:flash/message "Schedule deleted"]
                 [:schedules-list/get]]})
 
 (def events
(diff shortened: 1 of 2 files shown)


src/cljs/early_vote_site/early_vote_site_detail/events.cljs
```diff
@@ -198,7 +198,7 @@
 (defn delete-schedule-success
   [{:keys [db]} [_ early-vote-site-id]]
   {:db db
-   :dispatch-n [[:flash/message "Early vote site deleted"]
+   :dispatch-n [[:flash/message "Schedule deleted"]
```
lizstarin: Is the idea here that "Early vote site deleted" sounds like the site will no longer be an evs, ever?

tank157: Yeah, I think this was probably a copy-and-paste mistake. I noticed when I deleted a Schedule that the flash message said "Early vote site deleted" which made me nervous that it deleted the wrong thing. It did delete the Schedule, just the flash message was wrong.

lizstarin: 👍

PR #454, opened 2019-09-20
s/phantomjs/karma

tank157:
With phantomjs being deprecated/busted, it's time to upgrade to karma. And I upgraded the rest of the dependencies in the CLJS app while I was at it.


tank157:
@lizstarin This PR comes before the other one in this project, and just fixes the testing infrastructure so tests can be run again without having to locate phantomjs, which is no longer supported.

If you want to run this locally, you'll need to run `npm install` first, to get the dev dependencies downloaded.

lizstarin:
The app works and the tests run in my local environment. Looks good.

PR #456, opened 2019-09-23
Dasher elections

lizstarin:
CRUD routes for Elections, Early Vote Sites, Schedules, and Assignments now go through dasher. There's an issue or two left to resolve. In particular, date formatting: I can't figure out how to display dates in local time. There's also a (possibly related) bug? in the schedule form when it's in editing mode—the form doesn't display the current start and end times, just blank time input fields.


danny-pease:
closing this for now, can reopen if need be

PR #457, opened 2019-09-24
display of file options input depends on whether user is an admin

lizstarin:


tank157:
Hmmm, I was going to change the base branch to `staging` to deploy it there, but that introduced a lot more commits. This looks good, so I'll get it onto staging so we can accept it from there.

tank157:
I deployed this to staging, and now I can't get any uploads to work. I'm going to run this locally and see if I can reproduce the problem.

danny-pease:
closing this for now, can reopen if need be

PR #463, opened 2020-03-04
Update instructions

bmadhusu:
updated the instructions from my last run of Metis

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ you can run Metis locally to show a dashboard for that data.
 
 ### Prerequisites
 
+* XCode
 * [Node.js][node]
 * [Grunt][grunt]
     * Install Grunt CLI `npm install grunt-cli`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -139,12 +148,17 @@ AUTH0_DOMAIN_DASHBOARD=<auth0_domain> AUTH0_CLIENT_ID_DASHBOARD=<auth0_client_id
 AUTH0_AUDIENCE_DASHBOARD="some-audience-uri" ./node_modules/.bin/grunt default
 ```
 
-Grunt:
+So the options are:
+
+1. If using the `run-local` script, simply create a file `scripts/run-local` with the above contents and invoke with:
```
tank157: small change, the directory where we keep scripts is just called `script`, not plural like you have here and in the next line.

bmadhusu: Made a new commit; can you check now?

PR #465, opened 2020-03-20
EVS filters

lizstarin:
Filters by county and by evs type, per CO's request.
https://app.asana.com/0/0/1167401004717897/f

### DIFF
--- a/src/cljs/early_vote_site/election_detail/events.cljs
+++ b/src/cljs/early_vote_site/election_detail/events.cljs
@@ -62,9 +62,8 @@
                 :on-failure        [:get-early-vote-site-list/failure]}})
 
 (defn filter-early-vote-site-list
-  [{:keys [db]} [_ filter-type value]]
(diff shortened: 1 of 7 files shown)


src/cljs/early_vote_site/election_detail/events.cljs
```diff
@@ -60,6 +61,11 @@
                 :on-success        [:get-early-vote-site-list/success]
                 :on-failure        [:get-early-vote-site-list/failure]}})
 
+(defn filter-early-vote-site-list
```
tank157: Since this event handler is only altering the db, it doesn't need to be constructed like an `:fx` handler. You could do it simpler with the following function definition:
```
(defn filter-early-vote-site-list
  [db [_ filter-type value]
  (assoc-in db [:election-detail filter-type] value))
```

And then move it from the `:fx` collection of events to the `:db` collection.

PR #466, opened 2020-03-28
Schedule UI for EVS

lizstarin:


src/cljs/early_vote_site/utils.cljs
```diff
@@ -26,6 +27,18 @@
   (let [parsed (format/parse full-format date-str)]
     (coerce/to-date parsed)))
 
+(defn military-to-standard
```
tank157, line 30: Normally I'd think maybe we could find some functions in the cljs-time library to parse and then re-format the time, but I also know that dealing with times _without_ taking time zones and whatnot into account can be very tricky. Meaning it almost always wants to apply time zones to parsed and formatted times, rather than just dealing with them as pure clock times.

So, since we're not going to rely on libraries to handle this particular utility, perhaps some test cases to exercise this function would be appropriate?

lizstarin:
Added direct link to schedules form from the early vote site form. Added tests for `military-to-standard`.

PR #472, opened 2020-04-14
fixes edit evs > add schedule workflow

lizstarin:
There was a bug in a workflow I forgot about, to edit an Early Vote Site, then add Schedules.

### DIFF
--- a/src/cljs/early_vote_site/early_vote_site_form/events.cljs
+++ b/src/cljs/early_vote_site/early_vote_site_form/events.cljs
@@ -91,7 +91,7 @@
 
 (defn save-success [{:keys [db]} [_ add-schedules? response]]
   (let [form (:early-vote-site-form db)
-        evs-id (if (contains? form :id) (:id form) (second response))]
+        evs-id (get form :id (second response))]
     {:db (assoc db :early-vote-site-form db/fresh-early-vote-site-form)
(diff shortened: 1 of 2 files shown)


src/cljs/early_vote_site/early_vote_site_form/events.cljs
```diff
@@ -90,11 +90,13 @@
                   :on-failure          [:early-vote-site-save/failure]}}))
 
 (defn save-success [{:keys [db]} [_ add-schedules? response]]
-  {:db (assoc db :early-vote-site-form db/fresh-early-vote-site-form)
-   :dispatch-n [[:flash/message "Early Vote Site saved"]
-                (if add-schedules?
-                  [:navigate/early-vote-site-detail (second response)]
-                  [:navigate/election-detail])]})
+  (let [form (:early-vote-site-form db)
+        evs-id (if (contains? form :id) (:id form) (second response))]
```
tank157, line 94: I think this is fine, though as an alternative you could use the 3-arity form of `get` too:

```
(get form :id (second response))
```

It basically reads as "get the :id value out of the form hash, and if it's nil, return the second element of response instead." Honestly, both are fine and pretty easy to comprehend, just thought I'd mention that as an option if you haven't encountered it yet.

lizstarin: I hadn't!

PR #482, opened 2020-05-21
Edit schedule date format

lizstarin:
Fixes date display bug in the edit schedule form. https://app.asana.com/0/1157517224209433/1176802032772633/f

Fixes bug in edit schedule form that showed a start date as today, if it was earlier than today.


src/cljs/early_vote_site/early_vote_site_detail/views.cljs
```diff
@@ -44,7 +44,7 @@
                 :date-atom (:start-date-atom form)
                 :max-date-atom (:end-date-atom form)
                 :pikaday-attrs
-                {:min-date (js/Date.)
+                {:min-date (if editing? () (js/Date.))
```
gws, line 47: I did not know until this morning that you could write the empty list unquoted and not get a runtime error!

PR #483, opened 2020-05-29
Fixes a server crashing bug

tank157:
Occurs currently when someone submits a file upload without a type selected.

Product approved this approach to just set a default on the radio buttons.

### DIFF
--- a/metis@.service.template
+++ b/metis@.service.template
@@ -19,7 +19,7 @@ Environment=RUN_ENV=default
 
 ExecStartPre=-/usr/bin/docker kill ${CONTAINER}
 ExecStartPre=-/usr/bin/docker rm ${CONTAINER}
-ExecStartPre=/usr/bin/docker pull ${DOCKER_IMAGE}
+ExecStartPre=-/usr/bin/docker pull ${DOCKER_IMAGE}
 
 ExecStartPre=/bin/bash -c 'echo PORT=4000 > /tmp/${CONTAINER}--env'
 ExecStartPre=/bin/bash -c 'echo NODE_ENV="production" > /tmp/${CONTAINER}--env'
--- a/public/assets/js/app/controllers/data-upload/dataUploadController.js
+++ b/public/assets/js/app/controllers/data-upload/dataUploadController.js
@@ -13,6 +13,7 @@ function DataUploadCtrl($scope, $rootScope, Upload, $backendService, $route, $au
   // initialize page header variables
   $scope.setPageHeader("VIP Data Upload", breadcrumbs, "data-upload", "", null);
   $rootScope.isUploading = false;
+  $scope.type = "file"; //set a default value for the type radio button
   $scope.cannotSubmit = function() {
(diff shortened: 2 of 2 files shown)


tank157:
@bmadhusu Yeah, I debated whether or not to include it, and sort of reversed my stance on it for this particular repo only. Given that exceptions in the Node app seem to bubble all the way out to killing the docker container right now, and quay.io is of dubious stability, I think it's maybe safer to include the change to the fleet template. Previous to this bug I would have expected the node process to crash, but nodemon to restart it without it breaking out and crashing the docker container. But maybe that's just a dev setting?

Anyway, given that any server exception could take down the process, and any user trying it multiple times can bring down the dashboard, the tradeoff of having this in place for the node app is maybe worth it over the chance that quay.io could go down between us uploading an image and a server trying to download it and failing and silently running old code.

tank157:
All of which is to say I'm not super thrilled to include it, but the very real possibility of a user crashing the entire dashboard again with a different unknown bug is probably higher.

public/assets/js/app/controllers/data-upload/dataUploadController.js
```diff
@@ -13,6 +13,7 @@ function DataUploadCtrl($scope, $rootScope, Upload, $backendService, $route, $au
   // initialize page header variables
   $scope.setPageHeader("VIP Data Upload", breadcrumbs, "data-upload", "", null);
   $rootScope.isUploading = false;
+  $scope.type = "file"; //set a default value for the type radio button
```
lizstarin, line 16: My gut reaction is that it feels weird to have a validation that only exists in the UI, but realistically, I guess it's not like our dashboard users are _trying_ to crash our server.

tank157: Yeah, this is definitely in the "quick fix" category.

metis@.service.template
```diff
@@ -19,7 +19,7 @@ Environment=RUN_ENV=default
 
 ExecStartPre=-/usr/bin/docker kill ${CONTAINER}
 ExecStartPre=-/usr/bin/docker rm ${CONTAINER}
-ExecStartPre=/usr/bin/docker pull ${DOCKER_IMAGE}
+ExecStartPre=-/usr/bin/docker pull ${DOCKER_IMAGE}
```
gws, line 22: If this is a temporary change, do we want to throw a short comment in here reminding us to undo the `=-`?

tank157: The discussion on Friday was me sort of reversing course on not just including this change, but keeping it in place until we move over to k8s (and I just now realized the staging branch change I just made to mirror this is irrelevant because staging _is_ on k8s so changes to the fleet service file don't really matter...doh!). Anyway, given the failure mode for the Node express server is apparently killing the whole docker container, cross-referenced with the instability of quay.io lately, it seems safer to keep it in place for the time being so we don't have a user accidentally take down the site again with some other bug we don't know about.

PR #498, opened 2020-10-19
Vip 156 auth0 down prep

tank157:
Two components:

Printing out the full URL to download an error report in the Slack messages, to help us should Auth0 go down.

Paired with some work in data-processor, print out the checksum in Slack if it was included in a processing message.


alpha-dw (approved):
🔔

PR #505, opened 2021-03-12
Point build script towards new registry in us-west-2

jimkoglerdw:
This one is essentially identical to the one for `data-processor`

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,7 +5,7 @@ steps:
     - "region=us-west-2"
     branches: "staging"
     plugins:
-      - ecr#v2.1.1:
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -17,4 +17,4 @@ steps:
     plugins:
       - ecr#v2.1.1:
           login: true
-          region: "us-east-1"
\ No newline at end of file
+          region: "us-east-1"
```
gws, line 20: The label is a little misleading, the "deployment" step here is really just pushing something to ECR, and we want production images to go to `us-west-2` as well, so this does need to get updated and won't break anything.

tank157 (approved):
👍

PR #506, opened 2021-03-23
Vit 312 vit dashboard upgrade

tank157:
[Jira Card](https://democracyworks.atlassian.net/browse/VIP-312)

One of two PRs for this card, this updates the Dashboard code to use the new VIT. Needs to be paired with a PR in `gttp-relaunch` that allows the VIT to be configured with a Civic Info API Key that takes precedence over the one that was built into the deployment. This allows us to use the main production codebase, but override the Civic Info API Key to one that can access staged data. We can't use any of the staging deployments of the VIT as they are behind https basic auth challenges that prevent it from embedding properly.

### DIFF
--- a/public/vit.html
+++ b/public/vit.html
@@ -1,11 +1,12 @@
 <!doctype html>
 <html>
   <head>
-      <!-- VIT Staging Tool -->
-      <script type="text/javascript" src="https://tool-staging.votinginfoproject.org/app.js"></script>
+      <!-- VIT Source -->
+      <script type="text/javascript" src="https://votinginfotool.org/js/compiled/app.js"></script>
+      <link rel="stylesheet" href="https://votinginfotool.org/css/compiled/vit.css"/>
   </head>
 <body>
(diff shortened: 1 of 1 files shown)


public/vit.html
```diff
@@ -22,10 +23,11 @@
...
       if(apiKey) {
-        vit.load({key: apiKey,
-                  test: false,
-                  officialOnly: true,
-                  productionDataOnly: false});
+        vit.core.init("_vit", {"logo": {"type": "default"},
+                               "civic-info-api-key": apiKey,
+                               "official-only": true,
+                               "production-data-only": false,
+                               "display-mode": "desktop"});
```
lizstarin, line 30: Do we support the dashboard in mobile? Or is it just such an unlikely use case that we don't bother?

lizstarin (approved):
🚗 🛺 🚘

public/vit.html
tank157: No, the current dashboard doesn't really work in mobile form. Might be worth considering for the next version, though I suspect only a subset of functions would make sense in that context.

PR #511, opened 2021-06-23
Stop feed button

lizstarin:
Adds button and some UI for stopping a feed. Goes with[ a PR to `data-processor`. ](https://github.com/votinginfoproject/data-processor/pull/334)


pg/gets.js
```diff
@@ -32,6 +33,8 @@ module.exports = {
   getErrorsTotal: util.simpleQueryResponder(queries.errorsTotal, util.paramExtractor()),
   getApprovableStatus: util.simpleQueryResponder(queries.approvableStatus, util.paramExtractor()),
   approveFeed: util.simpleQueryResponder(queries.approveFeed, util.paramExtractor()),
+  stopFeed: util.simpleQueryResponder(queries.stopFeed, dashboardUtil.compoundParamExtractor([dashboardUtil.pathParamExtractor(['feedid']), // too hacky to use dashboard/util.js?
```
tank157, line 36: Using dashboard/util.js here seems ok.

pg/gets.js
```diff
@@ -32,6 +33,8 @@ module.exports = {
...
+                                                                                              dashboardUtil.bodyParamExtractor(['user_name'])])),
```
tank157, line 37: Ideally we wouldn't be taking in the username of who stopped the feed as a parameter, because it means a nefarious user could trick the application by making an authenticated request using another user's name. Instead we could pull the name from the user information after we've checked the authentication, because that can't be spoofed as easily.

For these purposes I'm not super concerned about this, but it would likely show up in a decent security audit as a problem. It's up to you if you'd like to re-work this a bit or not, but something to keep in mind for the re-write for sure.

I have a rough idea how we could make it work like this, but I admit I don't know what all we have access to once we've authenticated the user, so it could be we won't have the user_name value handily accessible.

tank157: I did a check, we don't have the user_name readily available after authentication, but we do have the user's id. So maybe we could leave the user_name as a parameter, but also include the user's id in the logs or something similar. That way if someone's ever like "I didn't stop this feed!" based on a Slack message, we could look up in the logs and find out if someone else did it instead.

tank157 (approved):
Generally looks good, some security concerns, but I feel like we could mitigate them with some extra logging (in the node express handler function, log both the user_name, but also use `getUserFromRequest` function from the authentication/services namespace to get the authenticated user and log their ID as well...ideally they will always match, but if someone raises an alarm that they didn't stop a feed, we can look at the logs for more info).

pg/gets.js
lizstarin: The username is just there as a string to show who made the request, though—it's not used for authentication via this request (unless there's something in your comment that I'm missing). Maybe `stop_requested` should be a datetime or boolean field instead, though?

tank157: I wouldn't worry about my concerns for this particular app. Just something to keep in mind with the re-write, that it's risky to trust certain values coming from even authenticated user browsers. In my mind, an authenticated user could still craft an HTTP request and replace their username with another username to make it look like a different user requested the stop. It'd be better to pull the username on the server side from the authenticated user info instead. But yeah, it's not a big deal here, just keep in mind in the next platform.

PR #520, opened 2021-08-04
allows additionalRoles in app_metadata

lizstarin:
Goes with this `dasher` PR: https://github.com/votinginfoproject/dasher/pull/27


public/assets/js/app/services/authService.js
```diff
@@ -174,6 +175,15 @@ vipApp.factory('$authService', function ($rootScope, $location, $timeout, $http,
     }
   };
 
+  function userToAdditionalRoles(metadata) {
+    if (metadata && metadata.additionalRoles) {
+      return metadata.additionalRoles;
+    } else {
+      console.log("no additional roles in metadata");
```
gws, line 182: I think this is a more "normal" condition than the lack of roles in the metadata (more like the `userToFips` above), so you can probably drop this log statement.

gws (commented):
Looks good, all the plumbing's there!

Am I missing how this ties into the application functionality (e.g. to enable an upload button or something)?

public/assets/js/app/services/authService.js
lizstarin: _Am I missing how this ties into the application functionality (e.g. to enable an upload button or something)?_

It doesn't, just thought it would be good to set it up. But now that you mention it, it might be good to hide the functionality when the user doesn't have permissions.

gws: Got it - sounds like there's no requirement to address it in this PR, should we surface this to Franklin or Anita for upcoming dashboard work?

gws (approved):
Lovely 📦🚢

PR #522, opened 2021-08-04
County user uploads

lizstarin:
Sets up use of `additionalRoles` in Metis; hides upload file choice radio buttons for users without upload permission.


gws (approved):
🙈 👀 🗒️

PR #525, opened 2021-08-09
removes conditions from form field

lizstarin:


gws (approved):
Wow, this is... not great behavior from Angular.

lizstarin:
It's also a suuuuuper old version, which is another reason we need to rebuild the dashboard.

PR #550, opened 2022-04-13
Vip 497 improve 403 messaging

dwxander:
Append message to error banner when feed upload forbidden


public/assets/js/app/app.js
```diff
@@ -353,9 +353,18 @@ vipApp.config(['$routeProvider', '$appProperties', '$httpProvider', '$logProvide
...
           }
 
+          // do not attempt to recover from forbidden responses
+          if (response.status === 403) {
+            return $q.reject(response);
+          }
+
+          // canRecover and responseOrNewPromsie are not defined elsewhere in the code
+          // and don't seem to do anything besides create errors - should this block
+          // be removed?
```
lizstarin, line 363: I think I changed my mind—let's get rid of this block (as well as the `if` block) and just use line 368. Because of the use case, I'm not so worried about what happens if for some reason this block is actually in use and something breaks.

dasher/services.js
```diff
@@ -65,7 +65,8 @@ function registerDasherServices(app) {
           res.status(500).send(error);
         } else if(response && response.statusCode){
           logger.info("response: ", response.statusCode);
-          res.status(response.statusCode).send();
+          logger.info("response body: ", body);
```
lizstarin, line 68: Did you mean to keep this log statement?

lizstarin (changes requested):
Almost there! My opinion has changed on that code block...

dasher/services.js
dwxander: I think I left it intentionally because the response status code was already being logged, but I can remove if it would be better

lizstarin: Either way. Just wanted to flag it.

lizstarin (approved):
🔥

PR #561, opened 2022-09-26
VIP-744: Add notify flag to optionally suppress dashboard emails

alice-dw:


alice-dw:
https://github.com/votinginfoproject/vip_admin/pull/326
https://github.com/votinginfoproject/data-processor/pull/373

notifications/sender.js
```diff
@@ -112,6 +112,10 @@ var badFips = function(fips) {
 
 var sendEmail = function(message, fips, contentFn) {
   authService.getUsersByFips(fips, function (users) {
+    logger.info(`notify flag: ${message.notify}`)
```
alice-dw, line 115: added some logging for the time being to make sure that
- message.notify is getting passed through at all
- it is it getting parsed as a boolean and not a string

danny-pease (approved):
📧

PR #562, opened 2022-09-28
VIP-744: Add notify type log

alice-dw:


alice-dw:
just adding a log and tagging for visibility. will remove once this ticket is done

PR #563, opened 2022-09-28
VIP-744: Log public Id when skipping email

alice-dw:


notifications/sender.js
```diff
@@ -112,9 +112,8 @@ var badFips = function(fips) {
 
 var sendEmail = function(message, fips, contentFn) {
   authService.getUsersByFips(fips, function (users) {
-    logger.info(`notify flag: ${message.notify}`)
-    logger.info(`typeof notify flag: ${typeof message.notify}`)
-    if (message.notify !== undefined && !message.notify) {
+    if (message.notify === false) {
+      logger.info(`Notify flag disabled, not sending email for message with publicId ${message.publicId}`);
```
alice-dw, line 116: this seems like the most sensible thing to log to me. open to suggestions otherwise

notifications/sender.js
```diff
@@ -112,9 +112,8 @@ var badFips = function(fips) {
 
 var sendEmail = function(message, fips, contentFn) {
   authService.getUsersByFips(fips, function (users) {
-    logger.info(`notify flag: ${message.notify}`)
-    logger.info(`typeof notify flag: ${typeof message.notify}`)
-    if (message.notify !== undefined && !message.notify) {
```
alice-dw, line 117: the deleted lines were confirming that message.notify gets correctly parsed as a bool, which it does, so logs are unnecessary and the condition can be simplified

alice-dw:
sorry one last one!!

notifications/sender.js
danny-pease: > this seems like the most sensible thing to log to me. open to suggestions otherwise

makes sense to me!

danny-pease (approved):
📬

PR #534, opened 2021-10-19
Vip 387 feed security fix

jimkoglerdw:
Creates a new middleware and applies it to one endpoint.

Next step to close this out is to apply it to any route with a :feedid parameter.

PR #565, opened 2022-10-04
VIP-762: Prevent users from viewing feeds from other jurisdictions

alice-dw:
Tested on /xml/overview-summary endpoint:
- `super-admin` can access all feeds
- guest user with FIPS `50` can access Vermont feeds
- guest user with FIPS `51` cannot access Vermont feeds
- guest user with FIPS `50` and `51` can access Vermont feeds


authentication/services.js
```diff
@@ -21,6 +24,35 @@ const checkJwt = jwt({
...
+  })();
+
+  conn.query(function(client, _) {
+    client.query(query, params, function(err, result) {
+      if (err) {
+        logger.error("Error querying for feeds");
+        logger.error(err);
+        res.writeHead(500, {'Content-Type': 'application/text'});
+        res.write("Server Error");
+        res.end();
```
alice-dw, line 44: this error handling is generic, it's done in most callbacks in this repo

authentication/services.js
```diff
@@ -21,6 +24,35 @@ const checkJwt = jwt({
...
+      } else {
+        requestedFeeds = result.rows.filter(feed => feed.public_id == req.params.feedid);
```
alice-dw, line 46: this is the main logic

alice-dw (commented):
see adjacent pr, only in staging: https://github.com/votinginfoproject/Metis/pull/534

i chose not to use the existing getFeeds functionality because this call doesn't really follow the simpleQueryResponder pattern since we need to use the result of the query before sending it in the response

pg/gets.js
```diff
@@ -16,7 +16,6 @@ var overviewTableRow = function(row, type, dbTable, link) {
 module.exports = {
   // Functions below return arrays for the various queries with the requirement of an ID.
   getFeeds: function(req, res) {
-		var roles = auth.getUserRoles(req);
```
alice-dw, line 19: random unused variable

dwxander (approved):
🔒

danny-pease (approved):
👾

PR #570, opened 2022-11-16
VIP-824: don't check feed id for non-auth routes

alice-dw:


alice-dw:
as of this PR, if a route should have the `checkUserAuthorizedForFeed` middleware if and only if it currently has the `checkJwt` middleware

dwxander (approved):

PR #574, opened 2022-12-17
VIP-842: Resolve Auth0 certificate issue

danny-pease:


gws:
The change in https://github.com/votinginfoproject/Metis/pull/574/commits/f764a292e19f574e6e86df51b5501e5bcc7566d5 to install a newer version of node-sass means we don't have to build it from source, which cuts about 2 minutes off of the build time.

The current code builds properly and should be ready to go to staging.

gws (approved):

PR #578, opened 2023-02-16
VIP 849 update ECR pipeline config

dwxander:
required for moving to DW buildkite


alice-dw (approved):
for my own understanding--where do these values come from? is that account id with the dw account or something?

dwxander:
oops sorry I meant to slack you to answer this - this is just the VIP account ID!

PR #582, opened 2023-03-01
VIP-828: Add auth checks for remaining routes

alice-dw:
Needed special handling because the way they were getting called wasn't including the authorization token from the browser on the request. Changed to override default link click behavior with an http.get request, which does include the token, then some special handling to download the report locally.


public/app/partials/5/aside.html
```diff
@@ -13,8 +13,7 @@ <h2>{{pageTitle}}<i class="fi-minus" ng-class="{'fi-plus': toggleAside}"><span c
       <br>
       {{errorCount.total_warnings | number}} Warnings in Feed
       <time id="feedDueDate" ng-if="getDueDateText(election.date)" class="due-date"><i class="fi-clock"></i> Final feed due {{getDueDateText(election.date)}}</time>
-      <a class="button button-download" href="{{errorReport}}">Download full error report</a>
-
+      <a class="button button-download" href="{{errorReport}}" ng-click="downloadReport($event)">Download full error report</a>
```
alice-dw, line 16: add special `ng-click` handler in html elements where reports are downloaded from

public/assets/js/app/app.js
```diff
@@ -798,7 +797,6 @@ vipApp.run(function ($rootScope, $appService, $location, $appProperties, $window
     return baseUrl + path;
   }
 
-  $rootScope.errorReportUrl = function() {
-    return '/db/feeds/' + $rootScope.feedId() + '/errors/report';
-  }
+  $rootScope.errorReportUrl = '/db/feeds/' + $rootScope.feedId() + '/errors/report';
```
dwxander, line 800: just for my own understanding, why do we want this to not be a function anymore/why was it one to begin with?

dwxander (approved):
🪪✅

public/assets/js/app/app.js
alice-dw: oops! good call. i changed it bc i wasnt sure but it actually doesn't work if it isn't a function, maybe because $rootScope.feedId() is undefined at the time we set this, so we call it later when $rootScope.feedId() is populated

danny-pease (approved):
🌻

PR #585, opened 2023-03-22
VIP-790: Add new non-paginated queries for feeds for authentication

alice-dw:


alice-dw:
context: when we make any db call in metis, we check to make sure the logged-in user is authorized to access a requested feed. 
we do that by
1. querying for the feeds a user is authorized to see, and then
2. searching those authorized feeds for the feed being requested. 

however, because of wonky pagination where the page number is not passed except during the initial feeds page load, the query in step 1 above was only returning the first 20 feeds for a user (aka feeds on page 1). therefore any requested feed from page 2+ would fail authorization because it would not be in the feeds that were returned by the query. 

solution:
- separate out a `feeds` query from a `paginatedFeeds` query, and same for `feedsForFipsCodes`. the `paginatedFeeds*` queries will return only the correct set of 20 feeds for a page, and the `feeds*` queries will return with no limit.
- we will call the `paginatedFeeds*` queries for loading the feeds page for the dashboard
- we will call the `feeds*` queries for authentication so that it is page-agnostic (don't have to track what page a feed was on as you click around to different pages in the feed)

potential drawback: each time we make a db call, we are querying all feeds that ever existed (worst case with a superuser). i checked the prod db and this is less than 3000 feeds, so i don't think there is a terrible performance drawback, but it's definitely worth considering whether we can pare down the number of feeds we have to look through to find the one being requested

pg/queries.js
```diff
@@ -31,16 +31,30 @@ module.exports = {
...
+        ORDER BY r.start_time DESC;",
+  paginatedFeeds: "SELECT DISTINCT ON (r.id) \
+                r.id, r.public_id, r.start_time, date(r.end_time) AS end_time, \
+                CASE WHEN r.end_time IS NOT NULL \
+                THEN r.end_time - r.start_time END AS duration, \
+                r.spec_version, r.complete, r.election_date, r.election_type, r.state, r.stop_requested \
+        FROM results r \
+        ORDER BY r.id DESC \
+        LIMIT 20 OFFSET ($1 * 20);",
+  paginatedFeedsForFipsCodes: "SELECT r.public_id, r.start_time, date(r.end_time) AS end_time, \
```
alice-dw, line 50: `paginatedFeeds` and `paginatedFeedsForFipsCodes` used to be named `feeds` and `feedsForFipsCodes`, and the sql is the same as it was. now, the queries named `feeds` and `feedsForFipsCodes` are what they were but without the limit

dwxander (approved):
📄

dwxander:
> potential drawback: each time we make a db call, we are querying all feeds that ever existed (worst case with a superuser). i checked the prod db and this is less than 3000 feeds, so i don't think there is a terrible performance drawback, but it's definitely worth considering whether we can pare down the number of feeds we have to look through to find the one being requested

this seems like a "we can fix it if it becomes a problem" type problem

